PR #47, opened 2022-08-10
Add/press2 75 Theme Installer (Wonder)

arunshenoy99:


officiallygod (approved):
Just make sure the add_to_queue method in PluginInstallTaskManager.php uses the self::$queue_name method to get the queue name. Rest Everything looks good to me and works fine.

PR #55, opened 2022-08-17
remove build from gitignore, use basic info data

arunshenoy99:


officiallygod (approved):

abhijitb (approved):

officiallygod:
We need to find a better way to update the options that are being changed and not run update_option every time, we can even think about segregating the data if the above is to be done.

PR #61, opened 2022-08-22
Add/PRESS2 93 Implement a API Queuing System

officiallygod:
The Current system makes an API request on every URL change and if the basic info step was triggered then the settings call is also done.

The New System has a Queue that is synced with the store and whenever a new step is loaded then the Calls to be made are queued into the queuer but before that, we need to flush out all the requests so that the queue is empty and all the previous items are saved into the DB.


arunshenoy99:
Closing as we have https://github.com/newfold-labs/wp-module-onboarding/pull/197 for this.

PR #62, opened 2022-08-22
Removed Default Values for site types

officiallygod:
Removed Default Values for Primary and Secondary Types


arunshenoy99:
Can we handle this in Javascript for our existing customers as well?

officiallygod:
> Can we handle this in Javascript for our existing customers as well?

Sure, I'll do that too.

arunshenoy99 (commented):
Should we be deleting the build files?

arunshenoy99 (approved):

PR #73, opened 2022-09-06
PRESS4-39 Fix responsiveness for ecommerce steps

aulisius:


arunshenoy99:
Screenshot 2022-09-07 at 11 40 03 AM
Screenshot 2022-09-07 at 11 42 52 AM
Screenshot 2022-09-07 at 11 43 31 AM
Screenshot 2022-09-07 at 11 43 44 AM
Screenshot 2022-09-07 at 11 44 05 AM
Screenshot 2022-09-07 at 11 44 48 AM
Screenshot 2022-09-07 at 11 44 57 AM
Screenshot 2022-09-07 at 11 46 02 AM
Looks way better than before, a few things I noticed
1. There are some warnings popping up when I build the assets. `npm run build` was the command I used.
2. We probably need to prevent content overflowing outside the card(on smaller screens and on opening console), not sure if the fix @abhijitb made will take care of it for these steps as well, please reach out to @abhijitb 
3. Step address seems to handle responsiveness better than the other 2 steps maybe we can use similar css? Attaching screenshots of how it looks on our breakpoint devices(some in portrait and some in landscape mode).

arunshenoy99:
The frontend checklist we came up with is something like this
1. Lower End: iPhone 4, portrait and landscape(Closest to 320px)
2. Higher End: 4k Monitor(3840px * 2160px)
3. Start from lower end to higher end using device type as responsive.

officiallygod:
Hey @aulisius, I found a few very small issues

Screenshot 2022-09-08 at 9 56 44 AM
Here the Text Boxes and Dropdown boxes are not aligned and it looks weird at first, maybe they can sorta fill the whole space? and be more uniform.


Screenshot 2022-09-08 at 9 57 00 AM
Here the Postal Code Input Box was overflowing for some reason for me.

Screenshot 2022-09-08 at 9 57 46 AM
The Minimum device we are targeting for our Module according to David is an iPhone 4
and I saw that the options on this page were quite not readable for a user out clear.
So the thing is the text overflows if the option is not selected but if it is then it is cut off.

Screenshot 2022-09-08 at 9 59 49 AM
Lastly again for an iPhone 4 the options were overflowing and were looking weird whereas they were aligned properly for the rest of the phones stacked vertically.

Thanks Please look into this.

abhijitb (approved):

PR #79, opened 2022-09-19
Theme Styles Selection Menu & Preview

arunshenoy99:
1. Live Preview: Theme Styles Menu Step.
2. Live Preview: Theme Styles Preview Step.
3. Dynamic injection of colors and typography steps based on user input in Preview Step.
4. Fix page refresh due to flow API call.


src/OnboardingSPA/pages/Steps/DesignThemeStyles/Preview/index.js
```diff
@@ -0,0 +1,206 @@
...
+			),
+			designSteps: orderBy(
+				designSteps.concat( steps ),
+				[ 'priority' ],
+				[ 'asc' ]
+			),
+		};
+	};
+
+	const removeColorAndTypographyRoutes = () => {
```
officiallygod, line 110: This dynamic step injection is really helpful and can be valuable later on too, Maybe we can source out the code in the reducer itself where we just pass the Step and it arranges it automatically on the basis of the priority, else to have such more page insertions we might have to duplicate this logic at multiple pages.

Say we have different types of Contact Us pages and the option can be shown if the user selects it in the Page Templates page, and if user doesn't select it we won't inject the page.

src/OnboardingSPA/pages/Steps/DesignThemeStyles/Preview/stylesheet.scss
```diff
@@ -0,0 +1,97 @@
...
+                display: inline-block;
+            }
+        }
+    }
+
+    &__live-preview-container {
+        position: relative;
+        width: 70%;
+        align-items: center;
+        border: 1px solid #e3dfdf;
```
officiallygod, line 88: Can you please add a `margin-bottom` also? The preview seems to be touching the bottom of the page and there is no clear end.

includes/Data/Patterns.php
```diff
@@ -0,0 +1,76 @@
...
+				'title'   => $pattern['title'],
+				'content' => self::cleanup_wp_grammar( $pattern['content'] ),
+				'name'    => $pattern['name'],
+			);
+		}
+
+		return false;
+	}
+
+	public static function get_theme_step_patterns_from_step( $step, $squash = false ) {
```
officiallygod, line 47: Perhaps a more generic name? 🙃 like the one used in **get_pattern_from_slug** Just a suggestion.

src/OnboardingSPA/pages/Steps/DesignThemeStyles/Preview/index.js
arunshenoy99: I thought about this at first but if you really think about it the priority of the steps is not really a store specific logic. The only job the action/reducer should be doing is to accept the data it is given and make sure it is updated in the store. It is just an `orderBy` function call that would be duplicated only in 2 cases
1. A new step is injected.
2. Priorities of one of the existing steps are changed
In all other cases (like removing steps) we would be doing an unnecessary `orderBy` operation cause the steps would already be in order.
So if any of our steps(consumers) do the above 2 operations it would be upto them to order these steps by where they expect them to be exactly like how our `routes/default-flow.js` is the one doing it instead of the store action/reducer. We can create separate reducers for `add` `remove` `modify` for each type of step object in the store etc but at this point it is just an overkill, once we move the drawer to a slot fill structure a number of these `topSteps` `designSteps` `getStartedSteps` would get removed from the store which might be right time to move to separate reducers.

includes/Data/Patterns.php
arunshenoy99: The function name describes exactly what it does, it returns `theme_step_patterns` which is our internal map of `themes->step->patterns` given a `step` identifier. It doesn't take a `slug` and return a `pattern`.

src/OnboardingSPA/pages/Steps/DesignThemeStyles/Preview/index.js
officiallygod: Actually having this in reducers is a good way to go forward, [LINK](https://redux.js.org/faq/code-structure/#how-should-i-split-my-logic-between-reducers-and-action-creators-where-should-my-business-logic-go:~:text=We%20recommend%20putting%20as%20much%20logic%20as%20possible%20into%20reducers.%20There%20are%20times%20when%20you%20may%20need%20some%20logic%20to%20help%20prepare%20what%20goes%20into%20the%20action%2C%20but%20reducers%20should%20do%20most%20of%20the%20work.)

I doubt in the near future if we see the page
https://www.figma.com/file/zajLn0Q74QSND5t7CIrpql/Modern-WordPress-Onboarding?node-id=35371%3A6297
If say a user selects About, in the next step he might be shown various About pages to select from and the orderBy can be reduced by having the logic in a way that cause the routes and allSteps needs to be updated in all cases
Rather passing an extra variable like 'Design' or 'Top' or 'Get Started' and subsequently running the updates for the specific case.

But yes this might be an Overkill rn if we are moving to Slot Fill soon.

officiallygod (approved):

PR #80, opened 2022-09-22
Add/press2 87 handle flow data dynamically

Yashita101:


arunshenoy99 (changes requested):
Pending bug fixes and requirements. Discussed with @Yashita101

includes/Services/FlowService.php
```diff
@@ -0,0 +1,157 @@
...
+					if (in_array($data, array_keys($value), true) ) {
+						unset($arrFlow[$key][$data]);
+					}
+					else {
+						$arrFlow[$key] = self::delete_wp_options_data_in_database( $data, $value );
+					}
+				}
+			}
+			return $arrFlow;
+	}
```
arunshenoy99, line 81: Can we combine these two into a single recursive function?

arunshenoy99 (changes requested):
I've not reviewed it thoroughly but found a few things that would help if we address them first. Will review again once these are fixed.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+	 * @var array
+	 * This is the key if a user tries to add a param externally
+	 */
+	protected static $mismatch_key = [];
+
+    public static function initalise_onboarding_flow_data() {
+        if ( ! ( $result = self::read_details_from_wp_options() ) ) {
+            $result = self::default_onboarding_data();
```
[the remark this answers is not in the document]
Yashita101: default_onboarding_data, now `get_default_flow_data` returns the flow_data, so the update and return is handled in that itself

includes/Services/FlowService.php
```diff
@@ -0,0 +1,153 @@
...
+		$flow_data_fixes = Flows::get_fixes();
+        if ( ! ( $flow_data = self::read_flow_data_from_wp_option() ) )
+            $flow_data = self::get_default_flow_data();
+		if ($flow_data_fixes) {
+			foreach($flow_data_fixes as $key=>$value) {
+				$flow_data = self::rename_keys($value, Flows::get_data(), $flow_data);
+				self::update_wp_options_data_in_database($flow_data);
+			}
+		}
+		$flow_data = self::update_flow_data($flow_data);
```
arunshenoy99, line 20: If flow data does not exist in the option why are we fixing the default flow data blueprint again?

includes/Services/FlowService.php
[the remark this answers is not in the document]
Yashita101: @arunshenoy99

arunshenoy99:
Discussed all the changes with @Yashita101

arunshenoy99:
Screenshot 2022-12-02 at 5 39 00 PM
The plugin broke when I switched to this branch and accessed onboarding.

includes/RestApi/FlowController.php
```diff
@@ -56,21 +57,30 @@ public function register_routes() {
...
+			if ($result) {
+				return new \WP_REST_Response(
+					$result,
+					200
+				);
+			}
+			return new \WP_Error(
+				'Flow Data does not Exist',
+				array( 'status' => 404 )
+			);
```
arunshenoy99, line 72:
```suggestion
		$result = FlowService::read_flow_data_from_wp_option();
		if( ! $result ) {
			$result = FlowService::get_default_flow_data();
		}
```

includes/RestApi/FlowController.php
```diff
@@ -79,7 +89,7 @@ public function get_onboarding_flow_data( \WP_REST_Request $request ) {
 	 * @return \WP_REST_Response|\WP_Error
 	 */
 	public function save_onboarding_flow_data( \WP_REST_Request $request ) {
-		$flow_data = array();
+		$default_flow_data = Flows::get_data();
```
arunshenoy99, line 92:
```suggestion
		$default_flow_data = FlowService::get_default_flow_data();
```
If I am not wrong this is the right function to call for getting the default flow data as it populates even the correct ecommerce values. Else I was getting publishing as the top priority for the commerce flow when it should be selling.

includes/RestApi/FlowController.php
```diff
@@ -90,30 +100,23 @@ public function save_onboarding_flow_data( \WP_REST_Request $request ) {
...
-
-		foreach ( $params as $key => $param ) {
-			if ( $value = $this->array_search_key( $key, $flow_data ) === false ) {
-				return new \WP_Error(
-					'wrong_param_provided',
-					"Wrong Parameter Provided : $key",
-					array( 'status' => 404 )
-				);
-			}
+		$flow_data = $this->get_wp_options_flow_data();
```
arunshenoy99, line 103:
```suggestion
```
I don't see anywhere this being used after retrieving.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+		$data['data']['topPriority']['priority1'] = 'selling';
+		$data['data']['siteType'] = array('label' => '', 'referTo' => 'business');
+		return $data;
+	}
+
+    /*
+	 * function to search for key in array recursively with case sensitive exact match
+	 */
+	public static function check_key_in_nested_array( $params, $flow_data, $header_key = 'Base Level' ) {
+		static $mismatch_key = [];
```
arunshenoy99, line 176:
```suggestion
		$mismatch_key = [];
```

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
+<?php
```
arunshenoy99, line 1: Screenshot 2022-12-05 at 9 41 49 PM
I vaguely remember discussing this but mentioning again in case we did not, phpcs and phpcbf both are failing on this file which means these will also fail in our release pipelines.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+namespace NewfoldLabs\WP\Module\Onboarding\Services;
+
+use NewfoldLabs\WP\Module\Onboarding\Data\Options;
+use NewfoldLabs\WP\Module\Onboarding\Data\Flows;
+use NewfoldLabs\WP\Module\Onboarding\Data\Data;
+
+class FlowService {
+
+	// An array to keep the Flow Data updated in the database
+	protected $updated_flow_data = array();
```
arunshenoy99, line 11:
```suggestion
```
Do not see `$this->updated_flow_data` being used anywhere. If you need this in any function it needs to be declared locally.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+
+	public static function initalize_flow_data() { 
+		$default_flow_data = Flows::get_data();
+        if ( ! ( $flow_data = self::read_flow_data_from_wp_option() ) )
+            return self::get_default_flow_data();
```
arunshenoy99, line 16:
```suggestion
        if ( ! ( $flow_data = self::read_flow_data_from_wp_option() ) )
            return self::get_default_flow_data();
        $default_flow_data = Flows::get_data();
```
Let's not read the default flow data until it is necessary.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+		$updated_flow_data = self::update_flow_data_recursive($default_flow_data, $flow_data, $updated_flow_data);
+		self::update_wp_options_data_in_database($updated_flow_data);
```
arunshenoy99, line 18: Database updates can fail so we probably need an `if` that checks whether this actually worked.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+        return $updated_flow_data;
+    }
+
+    private static function get_default_flow_data() {
```
arunshenoy99, line 22:
```suggestion
    public static function get_default_flow_data() {
```
No use making it private cause this can either ways be read via the other public functions.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+		if($flow_type == 'ecommerce') 
+			$flow_data = self::update_default_data_for_ecommerce( $flow_data );
+		self::update_wp_options_data_in_database( $flow_data );
+		return $flow_data;
+	}
+
+	/*
+	 * function to update the Flow Data (Blueprint) in an array recursively in comparison to Flows::get_data() (Database)
+	 */	
+	private static function update_flow_data_recursive(&$default_flow_data, &$flow_data, &$updated_flow_data) {
```
arunshenoy99, line 38:
```suggestion
	private static function update_flow_data_recursive(&$default_flow_data, &$flow_data, &$updated_flow_data) {
```
Any reason other than optimization to pass these by reference?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,191 @@
...
+class FlowService {
+
+	// An array to keep the Flow Data updated in the database
+	protected $updated_flow_data = array();
+
+	public static function initalize_flow_data() { 
+		$default_flow_data = Flows::get_data();
+        if ( ! ( $flow_data = self::read_flow_data_from_wp_option() ) )
+            return self::get_default_flow_data();
+		$updated_flow_data = self::update_flow_data_recursive($default_flow_data, $flow_data, $updated_flow_data);
```
arunshenoy99, line 17:
```suggestion
                 $updated_flow_data = array();
		$updated_flow_data = self::update_flow_data_recursive($default_flow_data, $flow_data, $updated_flow_data);
```
This will definitely break in lower PHP versions cause you are accessing `$updated_flow_data` before it even exists locally.

arunshenoy99 (commented):
This looks much better now Thanks! added a few more comments, will check `update_flow_data_recursive` and `update_post_call_data_recursive` once these are resolved cause those might change based on some of these comments.

includes/RestApi/FlowController.php
Yashita101: It is read and stored in the variable $flow_data, and then used in updating after the post call is made (line 114 )

includes/Services/FlowService.php
Yashita101: Yes, along with optimization, thee were cases, where the data was being overwritten, so to keep the checks in parallel, this was done

includes/Data/Flows.php
```diff
@@ -136,6 +136,28 @@ final class Flows {
...
 
+	/**
+	  * Has all the Flow Key parameters to be updated for the user in Key-Value pair.
+	  *
+	  * - OldKey is the existing key in $data which is to be renamed
+	  * - NewKey (mandatory) is the key to be replaced in the database and also to be manually 
+	  *   entered in $data instead of the value of 'OldKey' 
+	  * @return array
+	  */
+	  protected static $fixes = array(
```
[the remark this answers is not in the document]
Yashita101: Handled it by adding `retain_existing_value` in $fixes

includes/Services/FlowService.php
Yashita101: Handled it in WP_Admin.php

includes/Data/Flows.php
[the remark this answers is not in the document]
arunshenoy99: Thanks, Let's document the usage as well here?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,201 @@
...
+	public static function read_flow_data_from_wp_option() {
+		return \get_option( Options::get_option_name( 'flow' ), false );
+	}
+
+    /*
+	 * add onboarding flow options
+	 */
+	private static function save_details_to_wp_options( $data ) {
+		return \add_option( Options::get_option_name( 'flow' ), $data );
+	}
```
arunshenoy99, line 162:
```suggestion
```
More unused code?

includes/WP_Admin.php
```diff
@@ -119,6 +120,11 @@ public static function initialize() {
 			ThemeInstallTaskManager::queue_initial_installs();
 		}
 
+		if(FlowService::initalize_flow_data())
+			FlowService::initalize_flow_data();
+		else
+			FlowService::get_default_flow_data();
+
```
arunshenoy99, line 127: Not sure if this is gonna work cause the return value from `initalize_flow_data` is coming as `$updated_flow_data` which would always be there. To make sure this is saved in DB we probably need the return value from `self::update_wp_options_data_in_database($updated_flow_data);` in `initalize_flow_data`. Also not a good idea to call `initalize_flow_data` twice since these are costly operations.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+							else
+								$flow_data[$new_val['new_key']] = $default_flow_data[$new_val['new_key']];
+							unset($flow_data[$old_key]);
+						}	
+					}
+				}
+
+				// To update an existing value if the key exists in both, the blueprint and database
+				// All keys in the database and not in blueprint are not included
+				if (isset($flow_data) && array_key_exists($key, $flow_data) && !is_array($value)) {
```
arunshenoy99, line 59: Looking at the code I can think of a bug, I've tested and recorded a SC so I don't have to type the whole scenario out.
[screen-capture.webm](https://user-images.githubusercontent.com/38878906/206372600-fc21e861-5e05-475e-be14-071580534fba.webm)
Used example of `isFirstNFDOnboarding` but can easily happen for other values as well where we intentionally set it to a negative value like an empty array where default data does not have an empty array but the flow data in the DB does.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+				if ( is_array( $value ) ) {
+
+					// For Keys having Empty Arrays. Eg: isViewed, other
+					if(empty($value)) {
+						if(empty($flow_data[$key]))
+							$updated_flow_data[$key] = $value;
+						else
+							$updated_flow_data[$key] = $flow_data[$key];
+					}
+
```
arunshenoy99, line 82:
```suggestion
					if(empty($value)) {
					     $updated_flow_data[$key] = $flow_data[$key];
					}

```
Don't understand the logic behind this `if-else`, if `$value` is empty then we should just default to whatever `$flow_data[$key]` has, why check if `$flow_data[$key]` is also empty and then go ahead and assign it an empty value from `$value` when we know both are just empty arrays.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+						// To check if an Indexed Array is further Nested or Not
+						foreach($value as $index_key => $index_value) {
+
+							// For Indexed Arrays having values as an Array: Indexed/Associative
+							if(is_array($index_value)) {
+								if(empty($flow_data[$key][$index_key]))
+									$updated_flow_data[$key][$index_key] = $index_value;
+								else
+									$updated_flow_data[$key][$index_key] = $flow_data[$key][$index_key];
```
arunshenoy99, line 94: Should we not be going more deep here?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+			$flow_data = self::update_default_data_for_ecommerce( $flow_data );
+		self::update_wp_options_data_in_database( $flow_data );
+		return $flow_data;
+	}
+
+	/*
+	 * function to update the Flow Data (Blueprint) in an array recursively in comparison to Flows::get_data() (Database)
+	 */	
+	private static function update_flow_data_recursive(&$default_flow_data, &$flow_data, &$updated_flow_data) {
+		{
```
arunshenoy99, line 38:
```suggestion
	private static function update_flow_data_recursive(&$default_flow_data, &$flow_data, &$updated_flow_data) {
```
Double opening braces? 👀

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+			}
+			return $updated_flow_data;
+		  }
+	}
+
+	/*
+	 * function to update the Database recursively based on Values opted or entered by the User
+	 */	
+	public static function update_post_call_data_recursive(&$flow_data, &$updated_flow_data, &$params, &$flag = '') {
+		{
```
arunshenoy99, line 118:
```suggestion
	public static function update_post_call_data_recursive(&$flow_data, &$updated_flow_data, &$params, &$flag = '') {
```
Again double opening braces

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+			foreach ($flow_data as $key => $value)
+			{ 
+				if(is_array($params[$key]) && !is_array($value)) {
+					$flag = $key. ' => '. gettype($params[$key]) . '. Expected: ' . gettype($value);
+					break;
+				}
+
+				// Updates value entered by the user
+				if (isset($params[$key]) && array_key_exists($key, $params)) {
```
arunshenoy99, line 127:
```suggestion
				if (array_key_exists($key, $params) && isset($params[$key]) && ) {
```
We want to check if key exists first then check if it isset?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+						break;
+					}
+				}
+
+				elseif(isset($params) && !array_key_exists($key, $params)) {
+					$updated_flow_data[$key] = $flow_data[$key];
+				}
+
+				// Retains the Blueprint Value if no input from the User
+				else $updated_flow_data[$key] = $value;
```
arunshenoy99, line 141:
```suggestion
				// Retains the DB Value if no input from the User
				else $updated_flow_data[$key] = $value;
```
I might be wrong here but as far as I can see in relation to by previous comment, if the array_key does not exist in params or is not set or both, then we can just default to the value from the DB, don't know why we are checking `isset($params)` and then using `$flow_data[$key]` which is nothing but `$value`.

includes/Services/FlowService.php
arunshenoy99: This is still not resolved @Yashita101

includes/RestApi/FlowController.php
```diff
@@ -100,31 +97,32 @@ public function save_onboarding_flow_data( \WP_REST_Request $request ) {
...
-				);
-			}
+		$default_flow_data = FlowService::get_default_flow_data();
+		$mismatch_key = FlowService::check_key_in_nested_array($params, $default_flow_data);
+		if ( ! empty($mismatch_key) )  {
+			return new \WP_Error(
+				'wrong_param_provided',
+				"Wrong Parameter Provided",
+				array( 'status' => 404 , 'Mismatched Parameter(s)' => $mismatch_key)
+			);
```
arunshenoy99, line 119: Why are we searching flow data again, I thought lines 101-110 handled this?

includes/RestApi/FlowController.php
```diff
@@ -100,31 +97,32 @@ public function save_onboarding_flow_data( \WP_REST_Request $request ) {
...
 		}
 
-		$flow_data = array_replace_recursive( $flow_data, $params );
-
 		// update timestamp once data is updated
-		$flow_data['updatedAt'] = time();
+		$flow_data['updatedAt'] = strval(time());
```
arunshenoy99, line 123: Can we add a comment as to why this is made a string? And also should we updated `createdAt` to be a string as well?

arunshenoy99 (changes requested):
Let't run the php formatter with standards as Bluehost on all these files

includes/RestApi/FlowController.php
Yashita101: We are traversing through params in this case, here an error would be thrown if the user adds any key that does not exist in the blueprint, hence it needs to be handled separately

arunshenoy99: This needs to go above the post recursive call then like we discussed

includes/Services/FlowService.php
Yashita101: The check for $flow_data[$key] is required as the value turns out to be null for $other or $data keys, which are inside some array. Have modified it accordingly with appropriate checks.

includes/Services/FlowService.php
Yashita101: modified to $isset($params), it needs to be checked, otherwise it throws an error (especially in PHP v>8) if it receives null at any point of time during iteration.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+		{
+			$flow_data_fixes = Flows::get_fixes();
+
+			foreach ($default_flow_data as $key => $value)
+			{ 
+
+				// Any Key renamed is updated in the database with NewKey and the value from the OldKey is retained
+				if($flow_data_fixes) {
+					foreach ($flow_data_fixes as $old_key=>$new_val) {
+						if (isset($flow_data[$old_key]) && array_key_exists($new_val['new_key'], $default_flow_data)) {
```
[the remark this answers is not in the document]
Yashita101: Now checking if $flow_data is null or not, to avoid such a scenairo

includes/RestApi/FlowController.php
Yashita101: Changes have been made

includes/Services/FlowService.php
Yashita101: Yes, it is being handled inside the innermost array.

includes/RestApi/FlowController.php
arunshenoy99: I meant add a comment in the code as to why this is made a string.

includes/Services/FlowService.php
arunshenoy99: Please let's not mark a conversation as resolved until the other person agrees on the reply. Also the change you mentioned is not here.

arunshenoy99: Why check `isset($params)` when we have already accessed `$params[$key]` on line 119 i.e `is_array($params[$key])`, shouldn't it throw an error there itself if it receives null?

includes/RestApi/FlowController.php
Yashita101: As discussed, this change was reverted and the blueprint datatype for the same was made an integer

includes/Services/FlowService.php
Yashita101: This is taken care of After handling the conditional statements

includes/Services/FlowService.php
Yashita101: This would remain static, else it is not able to detect the mismatch key in the nested arrays

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+					} 
+
+					// To handle Associative Arrays gracefully
+					$updated_flow_data[$key] = self::update_flow_data_recursive($value, $flow_data[$key]);
+				}
+			}
+
+			// Adds or deletes key-value pairs in DB based on the modification in the blueprint
+			else
+				$updated_flow_data[$key] = $value;
```
arunshenoy99, line 90:
```suggestion
if( ! array_key_exists($key, $flow_data)) {
				$updated_flow_data[$key] = $value;
			}

				// Accepts the value of Exception List keys as is from the database OR non array OR empty array values
			if(isset($exception_list[$key]) || ! is_array($value) || sizeof($value) === 0)
				$updated_flow_data[$key] = $flow_data[$key];
				continue;

			// To handle Indexed Arrays gracefully
			if (count(array_filter(array_keys($value), 'is_string')) === 0) {
				// To check if an Indexed Array is further Nested or Not
				foreach($value as $index_key => $index_value) {
					// For Indexed Arrays having Non Associative Array Values
					(!is_array($index_value) && isset($flow_data[$key]))? 
						$updated_flow_data[$key] = $flow_data[$key] 
						: $updated_flow_data[$key] = $value;
				}
				continue;
			} 

			// To handle Associative Arrays gracefully
			$updated_flow_data[$key] = self::update_flow_data_recursive($value, $flow_data[$key]);
```
Let's always aim to make the code itself readable instead of adding comments that explain what it does

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+
+				else {
+					// To handle Indexed Arrays gracefully
+					if (count(array_filter(array_keys($value), 'is_string')) === 0) {
+						// To check if an Indexed Array is further Nested or Not
+						foreach($value as $index_key => $index_value) {
+							// For Indexed Arrays having Non Associative Array Values
+							(!is_array($index_value) && isset($flow_data[$key]))? 
+								$updated_flow_data[$key] = $flow_data[$key] 
+								: $updated_flow_data[$key] = $value;
```
arunshenoy99, line 78: If `$index_value` is an array we will probably need to do recursion again to go deep instead of just copying `$value` which is part of the reason why we cannot handle `palette`, `color` etc, I think since this has gone on for too long we can create a follow up story for the exception list and add this comment there.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+				$updated_flow_data[$key] = $value;
+		}
+		return $updated_flow_data;
+	}
+
+	/*
+	 * function to update the Database recursively based on Values opted or entered by the User
+	 */	
+	private static function update_post_call_data_recursive(&$flow_data, $params) {
+		static $mismatch_data_type = '';
```
arunshenoy99, line 99: We need to address this, `static` variables are often add unpredictable behaviour in recursion since they keep retaining their value, probably has to do with the way you are breaking and returning when there is a mismatch

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+		$exception_list = Flows::get_exception_list();
+
+		foreach ($flow_data as $key => $value)
+		{ 		
+			// Updates value entered by the user
+			if (array_key_exists($key, $params)){
```
arunshenoy99, line 105: Similar to my previous suggestion let's do a `!` and move the small logic in the `if` then `continue`

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+				}
+
+				// Error thrown if the datatype of the parameter does not match
+				if(strcmp(gettype($value), gettype($params[$key])) != 0) {
+					$mismatch_data_type = $key. ' => '. gettype($params[$key]) . '. Expected: ' . gettype($value);
+					break;
+				}
+
+				// Accepts non-Array Values entered by the user
+				elseif(!is_array($value))
```
arunshenoy99, line 120: Why `elseif` if you are using `break` in the `if` above?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+					$flow_data[$key] = $params[$key];
+
+				else {
```
arunshenoy99, line 123: Again no need of an else when we can always `continue`

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+					// To handle Indexed Arrays gracefully
+					if (count(array_filter(array_keys($params[$key]), 'is_string')) === 0 ) {
```
arunshenoy99, line 125: I see this logic being used in many places lets move it to a function with an appropriate name?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+
+				// Error thrown if the datatype of the parameter does not match
+				if(strcmp(gettype($value), gettype($params[$key])) != 0) {
+					$mismatch_data_type = $key. ' => '. gettype($params[$key]) . '. Expected: ' . gettype($value);
+					break;
+				}
+
+				// Accepts non-Array Values entered by the user
+				elseif(!is_array($value))
+					$flow_data[$key] = $params[$key];
```
arunshenoy99, line 121:
```suggestion
if (array_key_exists($key, $params)){			
				$flow_data[$key] = $value;
			}

			// Accepts the value of Exception List keys as is from the database
			if($exception_list && isset($exception_list[$key])) {
				$flow_data[$key] = $params[$key];
				continue;
			}

			// Error thrown if the datatype of the parameter does not match
			if(strcmp(gettype($value), gettype($params[$key])) != 0) {
				$mismatch_data_type = $key. ' => '. gettype($params[$key]) . '. Expected: ' . gettype($value);
				break;
			}

			// Accepts non-Array Values entered by the user
			if(!is_array($value)) {
				$flow_data[$key] = $params[$key];
				continue;
			}

			// To handle Indexed Arrays gracefully
			if (count(array_filter(array_keys($params[$key]), 'is_string')) === 0 ) {
				// If the Database value is empty or an indexed Array, to avoid Associative arrays to be overwritten (Eg: data)
				((sizeof($value) === 0) || (count(array_filter(array_keys($value), 'is_string')) === 0))?
					$flow_data[$key] = $params[$key]
					: $flow_data[$key] = $value;
				continue;
			}

			// To handle Associative Arrays gracefully
			$flow_data[$key] = self::update_post_call_data_recursive($value, $params[$key]);
										
```
Again, lets make the code itself readable instead of adding comments on most lines

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+
+				else {
+					// To handle Indexed Arrays gracefully
+					if (count(array_filter(array_keys($params[$key]), 'is_string')) === 0 ) {
+						// If the Database value is empty or an indexed Array, to avoid Associative arrays to be overwritten (Eg: data)
+						((sizeof($value) === 0) || (count(array_filter(array_keys($value), 'is_string')) === 0))?
+							$flow_data[$key] = $params[$key]
+							: $flow_data[$key] = $value;
```
arunshenoy99, line 129: Another place to go deep, lets add to the follow up story

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+								self::check_key_in_nested_array($value[$index_key], $flow_data[$key][$index_key], $key. " : " . $index_key);
+						}
+						continue;
+					}
+
+					// For Associative Arrays
+					self::check_key_in_nested_array($value, $flow_data[$key], $key);
+				}
+			}
+		}
```
arunshenoy99, line 196:
```suggestion
foreach($params as $key => $value){
			if(isset($exception_list[$key]) || ! is_array( $flow_data ) ) {
				continue;
			}

			// Error if the key added by the user is not present in the database
			if(!array_key_exists($key, $flow_data)) {
				$mismatch_key[]  = $header_key. " => " . $key;
				continue;
			}
			
			// To check sub-Arrays
			if ( ! is_array( $value ) || empty($value)) {
				continue;
			}

			// For Indexed Arrays
			if (count(array_filter(array_keys($value), 'is_string')) === 0) {
				foreach($value as $index_key=>$index_value) {
					// For Indexed Arrays having Associative Arrays as Values
					if(is_array($index_value))
						self::check_key_in_nested_array($value[$index_key], $flow_data[$key][$index_key], $key. " : " . $index_key);
				}
				continue;
			}

			// For Associative Arrays
			self::check_key_in_nested_array($value, $flow_data[$key], $key);
		}
```
Again, lets make the code itself readable

includes/Services/FlowService.php
arunshenoy99: Added code snippet for this

arunshenoy99 (changes requested):
Added code snippets this time so that you can implement faster, let me know if it is difficult to read/understand in that case I can just commit

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+		foreach ($default_flow_data as $key => $value)
+		{ 
+			// To align the datatype of the respective values with the one set in the blueprint
+			if(strcmp(gettype($value), gettype($flow_data[$key])) != 0) {
+				$updated_flow_data[$key] = $value;
+				continue;
+			}
+
+			// Any Key renamed is updated in the database with NewKey and the value from the OldKey is retained or not based on retain_existing_value
+			if($flow_data_fixes) {
```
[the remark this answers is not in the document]
Yashita101: Although it is hardcoded, I feel a check is required to avoid any kind of error. Shall add a better check like `sizeof`.

includes/Services/FlowService.php
Yashita101: Have removed that part of recursive code which was handling the indexed arrays having associative arrays, as these keys had more associative array values being added dynamically, which currently is not being handled as a scope of this story. This is in ToDo to eliminate the `$exception_list`

includes/Services/FlowService.php
Yashita101: The condition is the same, but the logic used in both the functions is different

includes/Services/FlowService.php
```diff
@@ -0,0 +1,199 @@
...
+				foreach ($flow_data_fixes as $old_key=>$replace_key) {
+					if (array_key_exists($old_key, $flow_data) && array_key_exists($replace_key['new_key'], $default_flow_data)) {
+						($replace_key['retain_existing_value'])?
+							$flow_data[$replace_key['new_key']] = $flow_data[$old_key]
+							: $flow_data[$replace_key['new_key']] = $default_flow_data[$replace_key['new_key']];
+						unset($flow_data[$old_key]);
+					}
```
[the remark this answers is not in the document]
Yashita101: Here, the correct value is being replaced in `$flow_data`, which is the database of the user. We are populating the updated data in `$updated_flow_data`, which is comparing and taking the values from `$default_flow_data` and `$flow_data` . Hence, `$flow_data` needs to be updated first with the right key-value from the fixes and then followed by populating the right key-values to `$updated-flow_data`

includes/Services/FlowService.php
Yashita101: Have modified the function accordingly, the changes above suggested were throwing errors for the values of the `$exception_list` keys, hence made changes as required.

includes/Services/FlowService.php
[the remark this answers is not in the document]
arunshenoy99: Exactly my point can't we just make the update to `$updated_flow_data` instead of `$flow_data` and skip the rest since this is also comparing with `$default_flow_data` on line 58?

includes/Services/FlowService.php
arunshenoy99: Let's create a story in the backlog for this

includes/Services/FlowService.php
arunshenoy99: Let me elaborate, can we move the duplicate logic `count(array_filter(array_keys($array), 'is_string')) === 0` into a function named `is_array_indexed` that accepts an array as the parameter, for ex here the function would accept `$params[$key]` as the param.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+						// For Indexed Arrays having Associative Arrays as Values
+						if(is_array($index_value))
+							self::check_key_in_nested_array($value[$index_key], $flow_data[$key][$index_key], $key. " : " . $index_key);
+					}
+					continue;
+				}
+
+				// For Associative Arrays
+				if( \is_wp_error(self::check_key_in_nested_array($value, $flow_data[$key], $key) )) {
+					return self::check_key_in_nested_array($value, $flow_data[$key], $key);
```
arunshenoy99, line 198: We don't need to call the function twice

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+					$flow_data[$key] = $params[$key]
+					: $flow_data[$key] = $value;
+				continue;
+			}
+
+			// To handle Associative Arrays gracefully
+			$flow_data[$key] = self::update_post_call_data_recursive($value, $params[$key]);
+			if( \is_wp_error($flow_data[$key]) ) {
+				return $flow_data[$key];
+			}
```
arunshenoy99, line 139: Let's not assign error to `$flow_data` since this is a referenced variable on line 95, let's assign only when it has a valid value.

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+		// update default data with ecommerce data
+		$data['data']['topPriority']['priority1'] = 'selling';
+		$data['data']['siteType']['referTo'] = 'business';
+		return $data;
+	}
+
+    /*
+	 * function to search for key in array recursively with case sensitive exact match
+	 */
+	public static function check_key_in_nested_array( $params, $flow_data, $header_key = 'Base Level' ) {
```
arunshenoy99, line 168: `question` What does the function return if there is no mismatch? The reason I am asking this is because if this is checking for mismatch then this can be named to something more appropriate like `find_mismatch`?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+				$flow_data[$key] = $params[$key];
+				continue;
+			}
+
+			// Error thrown if the datatype of the parameter does not match
+			if(strcmp(gettype($value), gettype($params[$key])) != 0) {
+				return new \WP_Error(
+					'wrong_param_type_provided',
+					"Wrong Parameter Type Provided : ". $key. ' => '. gettype($params[$key]) . '. Expected: ' . gettype($value),
+					array( 'status' => 404 )
```
arunshenoy99, line 116:
```suggestion
					array( 'status' => 400 )
```
404 is used if something is not found https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+	public static function check_key_in_nested_array( $params, $flow_data, $header_key = 'Base Level' ) {
+		$exception_list = Flows::get_exception_list();
+		foreach($params as $key => $value){
+			if(!isset($exception_list[$key]) && is_array($flow_data)) {
+				// Error if the key added by the user is not present in the database
+				if(!array_key_exists($key, $flow_data)) {
+					return new \WP_Error(
+						'wrong_param_provided',
+						"Wrong Parameter Provided",
+						array( 'status' => 404 , 'Mismatched Parameter(s)' => $header_key. " => " . $key)
```
arunshenoy99, line 177:
```suggestion
						array( 'status' => 400 , 'Mismatched Parameter(s)' => $header_key. " => " . $key)
```

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
...
+		$flow_data['createdAt'] = time();
+		// get current flow type
+		$flow_type = Data::current_flow();
+		// update default data if flow type is ecommerce
+		if($flow_type == 'ecommerce') 
+			$flow_data = self::update_default_data_for_ecommerce( $flow_data );
+		return $flow_data;
+	}
+
+	public static function update_flow_data($params) {
```
arunshenoy99, line 31:
```suggestion
	public static function get_updated_flow_data($params) {
```
Since it is not updating the DB

includes/Services/FlowService.php
```diff
@@ -0,0 +1,203 @@
+<?php
+namespace NewfoldLabs\WP\Module\Onboarding\Services;
+
+use NewfoldLabs\WP\Module\Onboarding\Data\Options;
+use NewfoldLabs\WP\Module\Onboarding\Data\Flows;
+use NewfoldLabs\WP\Module\Onboarding\Data\Data;
+
+class FlowService {
+
+	public static function initalize_flow_data() {
```
arunshenoy99, line 10: `Note for Later`: From now on Github actions will run a lint workflow on all our PHP files which also checks for docblocks, you can use https://marketplace.visualstudio.com/items?itemName=neilbrayfield.php-docblocker or https://marketplace.visualstudio.com/items?itemName=ronvanderheijden.phpdoc-generator to generate these. We can skip for now but lets follow up in flow data parts 2 and 3

includes/Services/FlowService.php
arunshenoy99: This is still unresolved

includes/Services/FlowService.php
```diff
@@ -0,0 +1,254 @@
...
+				if ( ! is_array( $value ) || empty( $value ) ) {
+					continue;
+				}
+
+				// For Indexed Arrays
+				if ( count( array_filter( array_keys( $value ), 'is_string' ) ) === 0 ) {
+					foreach ( $value as $index_key => $index_value ) {
+						// For Indexed Arrays having Associative Arrays as Values
+						if ( is_array( $index_value ) ) {
+							self::find_mismatch_key( $value[ $index_key ], $flow_data[ $key ][ $index_key ], $key . ' : ' . $index_key );
```
arunshenoy99, line 240: Are we missing a `return` here? Cause we have called the function but it does not do anything? I'm thinking we may not have caught this cause all these cases are in the exception list and are skipped?

includes/Services/FlowService.php
```diff
@@ -0,0 +1,254 @@
...
+		return $updated_flow_data;
+	}
+
+	/**
+	 * Function to update the Database recursively based on Values opted or entered by the User
+	 *
+	 * @param array $flow_data WP Options Data
+	 * @param array $params Params Data
+	 *
+	 * @return array
```
arunshenoy99, line 120:
```suggestion
	 * @return \WP_Error|array
```

includes/Services/FlowService.php
```diff
@@ -0,0 +1,259 @@
...
+			// Accepts non-Array Values entered by the user
+			if ( ! is_array( $value ) ) {
+				$flow_data[ $key ] = $params[ $key ];
+				continue;
+			}
+
+			// To handle Indexed Arrays gracefully
+			if ( self::is_array_indexed( $params[ $key ] ) ) {
+				// If the Database value is empty or an Indexed Array, to avoid Associative arrays to be overwritten (Eg: data)
+				( ( count( $value ) === 0 ) || ( count( array_filter( array_keys( $value ), 'is_string' ) ) === 0 ) ) ?
```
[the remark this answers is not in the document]
Yashita101: Yes, but only for the second condition

arunshenoy99 (approved):

officiallygod:
QA Done for:

1. Adding new key value pair which is added to blueprint, is there in the DB options    
2. Deleting existing key from DB which does not exist in Blueprint anymore    
3. Rename an existing key to new key and retain the old value ($fixes)    
4. Handle post request dynamically with user changes, and store user data
5. Get default data for a New User or if the DB is empty or does not exist
6. Retain the user entered value , even when the screen is refreshed, if any key is removed in the POST Call
7. Check for the type of value except for the ones in indexed array — mismatch data type
8. Error thrown if any key which is not in blueprint is added, except the ones in indexed array — mismatch key

PR #84, opened 2022-09-27
default state drop down value removed

prasanthreddy22:


abhijitb (approved):

arunshenoy99 (commented):
[screen-capture.webm](https://user-images.githubusercontent.com/38878906/193544910-35bfd848-6dd0-4aa5-90e3-3e127659dad6.webm)
@aulisius @prasanthreddy22 Hey guys I was testing this branch out and it seems like the previously selected `State` is not showing when I come back to the address step. In the recording I selected `Florida`/`Alaska` and went to the next step, but when I come back to the address step the state resets to blank again. Is this the behaviour we want?

arunshenoy99 (approved):

PR #86, opened 2022-09-29
Loading States for Ecommerce and Design Steps

arunshenoy99:


src/OnboardingSPA/components/Drawer/DrawerPanel/DesignThemeStylesPreview.js
```diff
@@ -14,36 +15,38 @@ const DesignThemeStylesPreview = () => {
...
-		setPattern( pattern?.body );
-		setGlobalStyles( globalStyles?.body );
-		let selectedStyle;
+		const patternResponse = await getPatterns(
+			currentStep.patternId,
+			true
+		);
+		const globalStylesResponse = await getGlobalStyles();
+		setPattern( patternResponse?.body );
+		setGlobalStyles( globalStylesResponse?.body );
```
officiallygod, line 40: Why don't we have checks like:
`if ( patternsResponse?.error ) {
			return updateThemeStatus( THEME_STATUS_NOT_ACTIVE );
		}
		const globalStylesResponse = await getGlobalStyles();
		if ( globalStylesResponse?.error ) {
			return updateThemeStatus( THEME_STATUS_NOT_ACTIVE );
		}` 
		
over here too? Or can't the function getting the responses itself do checks like this so as to prevent duplicacy in every step??

src/OnboardingSPA/components/ImageUploader/index.js
```diff
@@ -51,7 +51,7 @@ const ImageUploader = ({ icon, iconSetter }) => {
     function loader(){
         return (
         <div className="image-uploader_window">
-                <Loader/>
+                <ImageUploadLoader />
```
officiallygod, line 54: The only place this is being used is BasicInfo, let's just shift the whole thing to **Loaders/ImageUpload** ??

officiallygod:
I had this question, Say the Theme is Installed but not activated and so not in the queue itself, and say it does get activated in between, Then what happens?

officiallygod (approved):

PR #87, opened 2022-09-29
Fix/secondary sitetype css issue

abhijitb:


officiallygod:
Maybe we can pull in the trunk so as to remove the PRESS-4 changes and only keep our changes here.

arunshenoy99:
> Maybe we can pull in the trunk so as to remove the PRESS-4 changes and only keep our changes here.

Merged trunk into this branch.

PR #88, opened 2022-09-30
PRESS2-302 | Build Colors Design Step

officiallygod:


src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+					</div>
+					<div className='color-palette__name'>
+						{colorStyle?.charAt(0).toUpperCase() + colorStyle?.slice(1) }
+					</div>
+				</div>
+			);
+		}
+
+		return paletteRenderedList;
+	}
```
arunshenoy99, line 307: These can go into a ColorPaletteList component that accepts a list of color palettes.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+						onClick={(e) => selectCustomColor('secondary')}>
+						<div className={`custom-palette__below-row-icon ${secondaryColor && 'custom-palette__below-row-icon_selected_border'}`}
+							style={{ backgroundColor: `${secondaryColor ?? secondaryColorTemp}` }}>
+							{secondaryColor ? <>&#10003;</> : null}
+							</div>
+						<div className='custom-palette__below-row-text'>Secondary</div>
+					</div>
+					<div className='custom-palette__below-row'
+						onClick={(e) => selectCustomColor('tertiary')}>
+						<div className={`custom-palette__below-row-icon ${tertiaryColor && 'custom-palette__below-row-icon_selected_border'}`}
```
arunshenoy99, line 350: This can also go in a component maybe CustomColorPicker

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+
+	const { updatePreviewSettings, setCurrentOnboardingData } =
+		useDispatch(nfdOnboardingStore);
+
+	const colorPalettes = {
+		'calm': [
+			'#C7DBFF',
+			'#E6EBEE',
+			'#1A4733',
+		],
```
arunshenoy99, line 44:
```suggestion
	const colorPalettes = {
		'calm': {
			'primary': '#C7DBFF',
			'secondary': '#E6EBEE',
			'tertiary': '#1A4733',
		},
```
I think it is Ok to duplicate `primary, secondary, tertiary` as long as we can eliminate confusing indexes for selection. Also should this be coming via an API from the backend based on the active theme?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+							selectedThemeColorPalette[idx].color = colorPalettes[colorStyle][0];
+						break;
+					case 'background':
+						if (isCustomStyle && backgroundColorTemp)
+							selectedThemeColorPalette[idx].color = backgroundColorTemp;
+						else
+							selectedThemeColorPalette[idx].color = '#ffffff';
+						break;
+				}
+			}
```
arunshenoy99, line 115: We can eliminate such logic of using for loops and iteration variables if we can move to a more objective data structure.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+					case 'secondary':
+						if (colorPickerCalledBy == 'secondary' && secondaryColor)
+							selectedThemeColorPalette[idx].color = secondaryColor;
+						break;
+					case 'tertiary':
+						if (colorPickerCalledBy == 'tertiary' && tertiaryColor)
+							selectedThemeColorPalette[idx].color = tertiaryColor;
+						break;
+				}
+			}
```
arunshenoy99, line 200: We can use a better data structure that can eliminate these loops and idx variables

arunshenoy99 (changes requested):
Sorry for the late review, I've not tested this out yet as I'm still setting stuff up but I think there are quite a few changes to be done.

src/OnboardingSPA/pages/Steps/DesignColors/index.js
```diff
@@ -1,19 +1,97 @@
...
+				{pattern && (
+					<LivePreview
+						blockGrammer={pattern}
+						styling={'custom'}
+						viewportWidth={1300}
+					/>
+				)}
+			</div>
+		</CommonLayout>
+	);
```
arunshenoy99, line 94: Let's enclose this entire JSX inside the DesignStateHandler component that handles the installation of the theme required for the preview. https://github.com/newfold-labs/wp-module-onboarding/blob/release/v1.0.0/src/OnboardingSPA/components/StateHandlers/Design/index.js

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+	const [isLoaded, setIsLoaded] = useState(false);
+	const [globalStyles, setGlobalStyles] = useState();
+	const [selectedColors, setSelectedColors] = useState();
+	const [showColorPicker, setShowColorPicker] = useState(false);
+	const [isAccordionClosed, setIsAccordionClosed] = useState(true);
+
+	const [primaryColor, setPrimaryColor] = useState();
+	const [tertiaryColor, setTertiaryColor] = useState();
+	const [secondaryColor, setSecondaryColor] = useState();
+	const [backgroundColor, setBackgroundColor] = useState();
```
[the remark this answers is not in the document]
officiallygod: Combining them into a single object will then require de-structuring the object every time either of them changes and as per react best practice we should combine only states that do change together, also de-structuring can hinder performance as the color palette changes this value live and it has a lot of updates.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
-			</p>
+			</div>
+		);
+	}
+
+	return (
+		<div style={{ padding: '0 4px' }}>
+			<h2>{__('Color Palettes', 'wp-module-onboarding')}</h2>
+			{buildPalettes()}
+			{buildCustomPalette()}
```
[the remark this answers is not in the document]
officiallygod: They are very much specific to this page and can't even be reused in the typography page as that has a different outline, and they also use state variables due to which I insist on keeping them here. The in-place code helps to reduce the lag else we would have to use props and then manage every change.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+				"slug": "",
+       			"name": "",
+       			"color": [
+					{"slug": "primary", "name": "Primary", "color": ""},
+					{ "slug": "secondary", "name": "Secondary", "color": ""},
+					{ "slug": "tertiary", "name": "Tertiary", "color": ""},
+					{ "slug": "background", "name": "Background", "color": ""},
+				],
+				"supports": ["yith-wonder"]
+			};
```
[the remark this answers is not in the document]
officiallygod: Except for the supports part, everything is pretty much the same no matter which theme and there was an issue with flow data which prevented changing the design of the data structure considering the existing customers this was the way instead of changing it in Flows, also the colors page doesn't require any API calls of its own to get or set the data so I thought we are using 'yith-wonder' in many places so we can later use that from flows currentData itself when we start saving it, inside 

```
theme: {
  template: "",
  stylesheet: "",
  variation: ""
}
```

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+			setCurrentOnboardingData(currentData);
+		}
+		else {
+			selectedColors = currentData.data.palette[0];
+
+			if(selectedColors.slug === 'custom') {
+				setBackgroundColor(selectedColors?.color[3].color ?? null);
+				setPrimaryColor(selectedColors?.color[0].color ?? null);
+				setSecondaryColor(selectedColors?.color[1].color ?? null);
+				setTertiaryColor(selectedColors?.color[2].color ?? null);
```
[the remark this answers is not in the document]
officiallygod: I had to do this way cause the Flow saves it in a format that is easier for child theme generation and selectedColors contains the exact same data structure for saving in store and retrieving faster,

```
"color": [
	{"slug": "primary", "name": "Primary", "color": ""},
	{ "slug": "secondary", "name": "Secondary", "color": ""},
	{ "slug": "tertiary", "name": "Tertiary", "color": ""},
	{ "slug": "background", "name": "Background", "color": ""},
]
```
here the color is an array of objects due to which I cannot use color.primary

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,378 @@
...
+			}
+		} 
+		setSelectedColors(selectedColors);
+		setThemeColorPalette(currentData?.data?.palette[0]['slug'], selectedColors, selectedGlobalStyle);
```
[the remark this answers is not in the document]
officiallygod: Again this is because there is a bug in the Flows due to which I can't change the design of the data structure and I have to stick with the existing one.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: This is how the PreviewSettings is structured already and we have to do it this way.

includes/RestApi/Themes/ThemeColorsController.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Retrieves the active theme color variations.
+     *
+     * @return array|\WP_Error
+     */
+    public function get_theme_colors()
+    {
+        $active_theme = (\wp_get_theme())->get('TextDomain');
+        $pattern_slugs = self::$theme_colors[$active_theme];
```
arunshenoy99, line 113:
```suggestion
        $theme_color_palettes = get_fn
```

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,318 @@
...
-			</p>
+			</div>
+		);
+	}
+
+	return (
+		<div style={{ padding: '0 4px' }}>
+			<h2>{__('Color Palettes', 'wp-module-onboarding')}</h2>
+			{ colorPalettes && buildPalettes()}
+			{buildCustomPalette()}
```
arunshenoy99, line 315: Let's wait for `colorPalettes` in both the function calls? It looks like the `custom` color picker will render first then will get pushed down by the `theme` color palettes once the API returns them, this will look confusing on a mobile device.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,318 @@
...
+						/>
+					</Popover>
 				)}
-			</p>
+			</div>
+		);
+	}
+
+	return (
+		<div style={{ padding: '0 4px' }}>
```
arunshenoy99, line 312: Let's move this to css and use a proper parent base class to target. `theme-colors--drawer` or something like that. All the inner elements(the ones the functions render) should have that parent class as the prefix.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,318 @@
...
+			selectedColors?.color.forEach(color => {
+				selectedColorsLocalTemp[color.slug] = color.color;
+			});
+
+			setSelectedColorsLocal(selectedColorsLocalTemp);
+			return selectedColorsLocalTemp;
+		}
+	}
+
+	function LocalToState(selectedColorsLocalTemp, colorStyle) {
```
arunshenoy99, line 50:
```suggestion
	function localToState(selectedColorsLocal, colorStyle) {
```
Let's try removing `temp` as function argument names in any other places as well

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,318 @@
...
+		if (selectedColorsLocalTemp && colorStyle) {
+			let selectedColorsTemp = selectedColors;
```
arunshenoy99, line 52: I am still a bit confused as to why we are copying this into a temp variable, can you please elaborate?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,15 +1,318 @@
...
+			selectedColorsTemp.name = colorStyle?.charAt(0).toUpperCase() + colorStyle?.slice(1);
+
+			let colorsArray = [];
+			for (let colorName in selectedColorsLocalTemp) {
+				colorsArray.push({
+					'slug': colorName,
+					'name': colorName?.charAt(0).toUpperCase() + colorName?.slice(1),
+					'color': selectedColorsLocalTemp[colorName]
+				});
+			}
```
arunshenoy99, line 63: Let's use `map` here instead of creating a temp array and pushing into it.

arunshenoy99 (changes requested):

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: I had to resort to making an array cause that is how the data should be saved in DB and Store for easier child theme generation.

PR #91, opened 2022-10-12
Add/inject theme plugin preview assets when installed via cron

arunshenoy99:


officiallygod:
Hey @arunshenoy99! Please do check if it works fine without the flow too, cause for me it is stuck in an infinite reloads state.

[screen-capture.webm](https://user-images.githubusercontent.com/48691514/196966225-e260ca3e-4355-4d1f-9018-52a075ec8764.webm)

officiallygod (approved):

PR #102, opened 2022-10-28
PRESS2 331 Maintain State for colors and typography steps

officiallygod:


arunshenoy99 (commented):
Discussed with @officiallygod for a common parent component across all steps that sets up the preview state similar to DesignStateHandler. This will also affect the skeleton loader PR.

includes/Data/Options.php
```diff
@@ -38,6 +38,7 @@ final class Options {
 		'site_icon'                     => 'site_icon',
 		'show_on_front'                 => 'show_on_front',
 		'page_on_front'                 => 'page_on_front',
+		'custom_theme_styles'           => 'custom_theme_styles',
```
arunshenoy99, line 41:
```suggestion
		'theme_settings'           => 'theme_settings',
```

includes/RestApi/Themes/ThemeVariationsController.php
```diff
@@ -67,4 +84,25 @@ public function get_theme_variations( \WP_REST_Request $request ) {
...
 
+	/**
+	 * Saves the custom active theme variations.
+	 *
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public function set_theme_variation(\WP_REST_Request $request)
+	{
+		// The theme data with the new Colors and Fonts
+		$theme_data = json_decode($request->get_body(), true);
```
arunshenoy99, line 95: Should we validate this similar to flow data to ensure we do not introduce anything unnecessary?

includes/RestApi/Themes/ThemeVariationsController.php
```diff
@@ -43,16 +44,32 @@ public function register_routes() {
...
 
 	/**
 	 * Retrieves the active themes variations.
 	 *
-	 * @return \WP_REST_Response|\WP_Error
+	 * @return \array|\WP_Error
 	 */
 	public function get_theme_variations( \WP_REST_Request $request ) {
+
+		$default = $request->get_param('defaultValue');
```
arunshenoy99, line 63: Can you please explain the purpose of `defaultValue`?

src/OnboardingSPA/components/GlobalStyleParent/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Global Style Parent Component
+ * The Fetching of Global Style Object from either store or API is 
+ * common to a lot many places and this component does the trick import { useState, useEffect } from '@wordpress/element';for us.
+ *
+ * @returns Global Style Parent
+ */
+
+const GlobalStyleParent = ({ children }) => {
```
arunshenoy99, line 16:
```suggestion
const GlobalStylesProvider = ({ children }) => {
```
And this component needs to go inside the `LivePreview` components so it is easier to migrate to a new module in the future.

src/OnboardingSPA/components/GlobalStyleParent/index.js
```diff
@@ -0,0 +1,74 @@
...
+            setGlobalStyles({
+                ...selectedGlobalStyle,
+                'title': currentData.data.theme.variation,
+                'version': 2
+            });
+
+        updatePreviewSettings(
+            useGlobalStylesOutput(selectedGlobalStyle, storedPreviewSettings)
+        );
+        setIsLoaded(true);
```
arunshenoy99, line 62: Is there a reason to get all variations and then filter out the selected one, if I am not wrong a single API call will get us the current `settings` object from the option?

src/OnboardingSPA/components/GlobalStyleParent/index.js
```diff
@@ -0,0 +1,74 @@
...
+    const { updateThemeStatus, updatePreviewSettings } =
+        useDispatch(nfdOnboardingStore);
+
+    const getStylesAndPatterns = async () => {
+        const globalStyles = await getGlobalStyles();
+        if (globalStyles?.error) {
+            return updateThemeStatus(THEME_STATUS_NOT_ACTIVE);
+        }
+        let selectedGlobalStyle;
+        if (storedPreviewSettings?.title)
```
arunshenoy99, line 41: Not a good idea to rely on `title` better to check if it is an `object` that has `settings` and `styles`

src/OnboardingSPA/components/GlobalStyleParent/index.js
```diff
@@ -0,0 +1,74 @@
...
+            selectedGlobalStyle = storedPreviewSettings;
+        else if (currentData.data.theme.variation) {
+            selectedGlobalStyle = globalStyles.body.filter(
+                (globalStyle) =>
+                    globalStyle.title === currentData.data.theme.variation
+            )[0];
+        } else {
+            selectedGlobalStyle = globalStyles.body[0];
```
arunshenoy99, line 49: `body[0]` may not always contain the default style, better to tag default style by an ID in the backend API and then check for that here

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignTypography.js
```diff
@@ -75,20 +56,20 @@ const DesignTypography = () => {
...
-				'core/heading'
-			].typography.fontFamily;
-		setGlobalStyles( globalStylesCopy );
+
+		// globalStylesCopy.styles.typography.fontFamily =
+		// 	fontPalettesCopy[ fontStyle ]?.styles?.typography?.fontFamily;
+		// globalStylesCopy.styles.blocks[ 'core/heading' ].typography.fontFamily =
+		// 	fontPalettesCopy[ fontStyle ]?.styles.blocks[
+		// 		'core/heading'
+		// 	].typography.fontFamily;
```
arunshenoy99, line 72: We can remove this code?

arunshenoy99 (changes requested):

includes/RestApi/Themes/ThemeVariationsController.php
officiallygod: Cool let me do that.

includes/RestApi/Themes/ThemeVariationsController.php
officiallygod: So basically some pages do need the 6 variations like the Theme Styles page even though the user navigates back, so if DefaultValue is true the function behaves as it used to behave earlier before this story, else it checks if the option exists, and handles accordingly.

src/OnboardingSPA/components/GlobalStyleParent/index.js
officiallygod: Won't that increase our work as then every live preview component is gonna do this computation that is unnecessary as if this is a parent then it will only be done once.

src/OnboardingSPA/components/GlobalStyleParent/index.js
officiallygod: So basically that won't happen unless we are on colors page, till then this will act like before and throw 6 variations, but later just one so iterating later also wont be a task as the first one will match.

src/OnboardingSPA/components/GlobalStyleParent/index.js
officiallygod: Sure but i found that styles and GlobalStyles were used so can't rely on them too, sure will add the settings check too.

src/OnboardingSPA/components/GlobalStyleParent/index.js
officiallygod: Sure Will add that.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignTypography.js
officiallygod: So this is the part which will work once the use-global-styles is fixed to work like before, I can sure go ahead and remove this if you want.

includes/RestApi/Themes/ThemeVariationsController.php
arunshenoy99: The validation can move inside a `validate_callback` function as part of the endpoint registration https://www.shawnhooper.ca/2017/02/15/wp-rest-secrets-found-reading-core-code/

src/OnboardingSPA/utils/api/themes.js
```diff
@@ -18,9 +18,19 @@ const init = () => {
 	} );
 };
 
-const getGlobalStyles = async () => {
+const getGlobalStyles = async ( defaultValue = false ) => {
```
arunshenoy99, line 21:
```suggestion
const getGlobalStyles = async ( variations = false ) => {
```

arunshenoy99 (changes requested):
Looks good! A couple minor refactors and let's run formatters on the PHP and JS files

includes/RestApi/Themes/ThemeVariationsController.php
officiallygod: Done

src/OnboardingSPA/utils/api/themes.js
officiallygod: Done

arunshenoy99 (approved):

PR #107, opened 2022-11-29
PRESS2 347 Build Site Features MVP Step

officiallygod:


includes/Services/PluginUninstaller.php
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		// Removes directory and files of a plugin
+		if ( ! \delete_plugins( array( $plugin_path ) ) ) {
+			return new \WP_Error(
+				'nfd_onboarding_error',
+				'Unable to Delete the Plugin',
+				array( 'status' => 500 )
+			);
+		}
```
arunshenoy99, line 144: https://developer.wordpress.org/reference/functions/delete_plugins/ looks like we probably need to check for `WP_Error`

includes/TaskManagers/PluginUninstallTaskManager.php
```diff
@@ -0,0 +1,145 @@
...
+	function __construct() {
+		// Ensure there is a thirty second option in the cron schedules
+		add_filter( 'cron_schedules', array( $this, 'add_thirty_seconds_schedule' ) );
+
+		// Thirty second cron hook
+		add_action( 'nfd_module_onboarding_plugin_uninstall_cron', array( $this, 'uninstall' ) );
+
+		// Register the cron task
+		if ( ! wp_next_scheduled( 'nfd_module_onboarding_plugin_uninstall_cron' ) ) {
+			wp_schedule_event( time(), 'thirty_seconds', 'nfd_module_onboarding_plugin_uninstall_cron' );
```
arunshenoy99, line 31: Can uninstalls be faster than `30 seconds`?

includes/TaskManagers/PluginUninstallTaskManager.php
```diff
@@ -0,0 +1,145 @@
...
+
+		// Recreate the PluginInstall task from the associative array.
+		$plugin_uninstall_task = new PluginUninstallTask(
+			$plugin_to_uninstall['slug'],
+			$plugin_to_uninstall['priority'],
+			$plugin_to_uninstall['retries']
+		);
+
+		// Update status to the current slug being installed.
+		\update_option( Options::get_option_name( 'plugins_uninit_status' ), $plugin_uninstall_task->get_slug() );
```
arunshenoy99, line 78: In reference to my first comment we can probably remove `plugin_uninit_status`

includes/TaskManagers/TaskManager.php
```diff
@@ -7,6 +7,7 @@ final class TaskManager {
 
 	 protected $task_managers = array(
 		 'NewfoldLabs\\WP\Module\\Onboarding\\TaskManagers\\PluginInstallTaskManager',
+		 'NewfoldLabs\\WP\Module\\Onboarding\\TaskManagers\\PluginUninstallTaskManager',
```
arunshenoy99, line 10: Let's combine these `TaskManagers` into a single `PluginsManager` so that we are not creating a new task manager for every single operation. Long term the goal is to simply remove all these and move to a single `TaskManager`

src/OnboardingSPA/components/CheckboxTemplate/CheckboxItem/index.js
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @param {string} icon     - The icon name of the Item
+ * @param {string} title    - The Main Title of the Item
+ * @param {string} subtitle - The Sub Title of the Item
+ * @param {string} desc     - The Description of the Item
+ *
+ * @return CheckboxItem
+ */
+
+const CheckboxItem = ( {
```
arunshenoy99, line 18: Probably a bad idea to call it `CheckboxItem` and then pass it something that is very specific to site features which is `slug`. This also needs to take a `className` for custom CSS

src/OnboardingSPA/components/CheckboxTemplate/CheckboxList/index.js
```diff
@@ -0,0 +1,50 @@
...
+ * This returns a List of Checkbox Items to be placed dynamically on screen
+ *
+ * @param          customPluginsList.callback
+ * @param {Object} customPluginsList                   - The List to be shown with a Title, Subtitle and a Description
+ *
+ * @param          customPluginsList.selectedPlugins
+ * @param          customPluginsList.customPluginsList
+ * @return CheckboxList
+ */
+const CheckboxList = ( { callback, selectedPlugins, customPluginsList } ) => {
```
arunshenoy99, line 14: Again this is very specific to siteFeatures when you use `selectedPlugins` and `customPluginsList`. I believe the same kind of component will be used for one of the ecommerce steps so let's please make this and the `CheckboxItem` more generic so that it can support that as well

src/OnboardingSPA/utils/api/plugins.js
```diff
@@ -30,3 +30,11 @@ export const getPluginStatus = async ( plugin ) => {
 		} )
 	);
 };
+
+export const getCustomPluginsList = async () => {
```
arunshenoy99, line 34:
```suggestion
export const getSiteFeatures = async () => {
```

src/OnboardingSPA/utils/api/plugins.js
```diff
@@ -30,3 +30,11 @@ export const getPluginStatus = async ( plugin ) => {
...
+	return await resolve(
+		apiFetch( {
+			url: onboardingRestURL( 'plugins/features' ),
```
arunshenoy99, line 37:
```suggestion
			url: onboardingRestURL( 'plugins/site-features' ),
```

includes/Data/SiteFeatures.php
```diff
@@ -0,0 +1,129 @@
...
+		$idx                  = 0;
+		$default_plugins_hash = array();
+
+		foreach ( self::$default_plugins as $default_plugin ) {
+			$default_plugins_hash[ $default_plugin['slug'] ] = $idx;
+			$idx += 1;
+		}
+
+		return $default_plugins_hash;
+	}
```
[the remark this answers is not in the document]
officiallygod: Yeah I was initially doing that but then the front end would have to be expecting a key value Object rather than a array and then the further use of that component will require this exact signature.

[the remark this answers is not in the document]
officiallygod: Also, we won't be able to use **map** and all further in our Components

includes/RestApi/PluginsController.php
```diff
@@ -242,4 +278,59 @@ public function get_status( \WP_REST_Request $request ) {
...
+
+		  $plugin_body = json_decode( $request->get_body(), true );
+		  $plugins     = isset( $plugin_body['plugins'] ) ? $plugin_body['plugins'] : false;
+
+		if ( $plugins ) {
+			foreach ( $plugins as $plugin => $decision ) {
+				if ( $decision ) {
+					PluginInstallTaskManager::add_to_queue(
+						new PluginInstallTask(
+							$plugin,
```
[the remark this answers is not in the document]
officiallygod: I did not quite understand this? 
**get_uninstall_plugin_args** does check if plugins is provided as param

includes/RestApi/PluginsController.php
```diff
@@ -242,4 +278,59 @@ public function get_status( \WP_REST_Request $request ) {
...
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public function uninstall( \WP_REST_Request $request ) {
+
+		  $plugin_body = json_decode( $request->get_body(), true );
+		  $plugins     = isset( $plugin_body['plugins'] ) ? $plugin_body['plugins'] : false;
+
+		if ( $plugins ) {
+			foreach ( $plugins as $plugin => $decision ) {
+				if ( $decision ) {
```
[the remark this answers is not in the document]
officiallygod: Using the decision we need to either Install or Uninstall so the logic for that should be in a much more generic place and this logic was very much specific to the site features POST request.

includes/Services/PluginUninstaller.php
```diff
@@ -0,0 +1,149 @@
...
+/**
+ * Class PluginUninstaller
+ * This class is responsible to Uninstall a specified plugin
+ */
+
+class PluginUninstaller {
+
+	public static function uninstall( $plugin ) {
+
+		$plugin_list = Plugins::get_squashed();
```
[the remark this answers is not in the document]
officiallygod: Again these plugins can be uninstalled only if a path exists and the path itself is derived from the install list due to which the plugin path itself will never come up or anything other than the ones in the install init list will never be uninstalled.

includes/Services/PluginUninstaller.php
officiallygod: Incase of an error we do pass forward that error to the Uninstall Task Manager itself which increments the count and tries again.

includes/TaskManagers/PluginUninstallTaskManager.php
officiallygod: Yes, it can be but to be on the safer side I choose this as sometimes what if plugins are huge in size as in they save a lot of user data? That can take some time to wipe off the disk.

includes/RestApi/PluginsController.php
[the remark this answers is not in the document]
arunshenoy99: An API named `uninstall` is queuing plugins for install right now.

So the architecture I was thinking of is something like this
2. We can do `PluginUninstallTaskManager::uninstall()` in the `uninstall` API, the TaskManager's uninstall can decide whether to un-queue or uninstall. If not, I am also fine with keeping the logic in the API as well.
3. Point 2 is the only thing the `uninstall` API should be doing.
4. Since we already have a `GET site-features` endpoint we can create one more for handling `site-features` which can make the `decision` of calling the `install` or the `uninstall` API's internally?

We create a new story to handle the `long-term` approach for `site-features` and `plugin_init` mappings based on flow/plan.

arunshenoy99 (approved):

PR #111, opened 2022-12-01
Onboarding Completion - Generate Child Theme, Publish Site Pages

arunshenoy99:
1. Generate Child Theme
2. Set homepage and publish all site page selections


src/OnboardingSPA/pages/Steps/SitePages/index.js
```diff
@@ -81,28 +83,52 @@ const StepSitePages = () => {
...
 			}
 		}
 		setIsLoaded( true );
 	};
 
-	const handleCheckedPages = ( selectedPages ) => {
+	const stateToFlowData = ( selectedPages, sitePages ) => {
+		return sitePages !== false
+			? sitePages
+					?.filter( ( sitePage ) => {
```
officiallygod, line 96: Filter and Map are gonna first iterate the array once then the selected list is iterated again i.e. 4 + 3 for the site pages step, maybe you can try something like reduce? (This would do that in 4)

```suggestion
					var checkedPagesCopy = sitePagesResponse.body.reduce(function (result, sitePage) {
						if (sitePage.selected)
							result.push( sitePage.slug );
						return result;
					}, []);
```

https://stackoverflow.com/questions/34398279/map-and-filter-an-array-at-the-same-time#:~:text=You%20should%20use%20Array.reduce%20for%20this.

src/OnboardingSPA/pages/Steps/Complete/index.js
```diff
@@ -0,0 +1,57 @@
...
+	const { nextStep, brandName } = useSelect( ( select ) => {
+		return {
+			nextStep: select( nfdOnboardingStore ).getNextStep(),
+			brandName: select( nfdOnboardingStore ).getNewfoldBrandName(),
+		};
+	}, [] );
+
+	const contents = getContents( brandName );
+
+	useEffect( async () => {
```
officiallygod, line 28: Maybe take a look at this? Maybe we can factor this out into a function? Just in Case

https://ultimatecourses.com/blog/using-async-await-inside-react-use-effect-hook#:~:text=This%20WORKS%2C%20but%20you%20should%20avoid%20it.%20Why%3F%20Because%20React%E2%80%99s%20useEffect%20hook%20expects%20a%20cleanup%20function%20returned%20from%20it%20which%20is%20called%20when%20the%20component%20unmounts.%20Using%20an%20async%20function%20here%20will%20cause%20a%20bug%20as%20the%20cleanup%20function%20will%20never%20get%20called.%20Yikes!%20So%20what%20do%20we%20do%3F

src/OnboardingSPA/pages/Steps/Complete/contents.js
```diff
@@ -0,0 +1,44 @@
...
+			subtitle: __(
+				'We’re assembling your unique design and installing useful tools',
+				'wp-module-onboarding'
+			),
+		},
+		errorState: {
+			title: sprintf(
+				/* translators: 1: Brand 2: site */
+				__(
+					'Hang tight, we’re building your %s %s',
```
officiallygod, line 26: Maybe we can ask David about these error page messages? As they are quite misleading where as this is a still page and nothing is done in background.

src/OnboardingSPA/data/routes/default-flow.js
```diff
@@ -615,7 +627,8 @@ export const initialTopSteps = () => {
 	const topSteps = filter( steps, ( step ) => {
 		return (
 			! step.path.includes( '/step/get-started' ) &&
-			! step.path.includes( '/step/design' )
+			! step.path.includes( '/step/design' ) &&
+			! step.path.includes( '/step/complete' )
```
officiallygod, line 631: Won't it be better to have a boolean or some sort while describing a page up above rather than again and again adding such hardcoded texts? Just thinking out loud.
`isTopPage: True or False`

includes/Data/Themes.php
```diff
@@ -79,7 +85,7 @@ private static function check_approved( $value ) {
 	  * @return array
 	  */
 	public static function step_preview_data() {
-		$theme_step_data   = Patterns::get_count_of_patterns();
+		$theme_step_data = Patterns::get_count_of_patterns();
```
officiallygod, line 88: Hey Sorry, but on Line 74 is self::$wp_slugs present in this file?

includes/RestApi/Themes/ThemeGeneratorController.php
```diff
@@ -45,17 +47,25 @@ public function register_routes() {
...
+		$flow_data_option = \get_option( Options::get_option_name( 'flow' ), false );
+		if ( $flow_data_option === false || ! isset( $flow_data_option['data'] ) ) {
+			return new \WP_Error(
+				'nfd_onboarding_error',
+				'Flow data does not exist to generate a child theme.',
+				array( 'status' => 500 )
+			);
+		}
+
+		$flow_data = $this->validate_and_sanitize_flow_data( $flow_data_option['data'] );
```
officiallygod, line 59: Had a doubt that as this function sets the $flow_data['theme']['template'] attribute should that be not saved in flow data?

src/OnboardingSPA/data/routes/default-flow.js
arunshenoy99: I like the thinking but the problem is not so simple as we have "fake-steps" being injected in cases like `design` and `get-started` which `complete` does not need. This also means that `design` and `get-started` are "fake" top steps which simply do not have a component to render so `isTopPage` would not exactly make sense. I would say it is a good idea but something to do once we see some general pattern in this filter.

includes/Data/Themes.php
arunshenoy99: Nopeee, great catch! can't believe this lived on for so long, all the more reason to do code reviews earlier.

src/OnboardingSPA/pages/Steps/Complete/index.js
arunshenoy99: Ohhhh Thanks again! I will keep this in mind and also create stories for wherever we may have missed doing this.

src/OnboardingSPA/pages/Steps/Complete/contents.js
arunshenoy99: It does say an error occurred below the "installing/building" so I believe we should be fine with this and if I am not wrong this came from the Bluerock pattern which also shows something similar to this. But will bring it up.

officiallygod (approved):

officiallygod:
Hey there on Themes.php line 25 we have a mapping for **default** and **ecommerce**,
but does Data::current_flow(); return default or wp-setup??

PR #116, opened 2022-12-08
PRESS 2 359 Call Site Features API in the Theme Generation interstitial state

officiallygod:
This Calls the Site Features API to queue/enqueue all the Installs and Uninstalls of various plugins according to the choices made by the user.

### DIFF
--- a/includes/Services/PluginUninstaller.php
+++ b/includes/Services/PluginUninstaller.php
@@ -13,13 +13,6 @@ class PluginUninstaller {
 
 	public static function uninstall( $plugin ) {
 
-		$position_in_queue = PluginInstallTaskManager::status( $plugin );
-		if ( $position_in_queue !== false && $position_in_queue !== 0 ) {
-			PluginInstallTaskManager::remove_from_queue(
-				$plugin,
-			);
-		}
-
 		$plugin_list = Plugins::get_squashed();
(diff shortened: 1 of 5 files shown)


arunshenoy99 (changes requested):
[screen-capture.webm](https://user-images.githubusercontent.com/38878906/206671828-fcf9b6cb-a067-4eab-ba55-bbc8f3ec7c07.webm)
Attaching SC, whenever I click next after site-features I immediately get what-next or get a blank what-next page but never get the loading state. The API calls work which probably means we need to await on them to complete.

officiallygod:
> 

Ohhooo Lemme see that!

arunshenoy99 (commented):
A few things I found, I de-selected jetpack from the list when it was still in the `install_queue` but it was still added to the `uninstall_queue`, we should never add to the `uninstall_queue` if it was never installed in the first place, attaching SS
Screenshot 2022-12-12 at 11 07 45 AM

Screenshot 2022-12-12 at 11 07 33 AM

arunshenoy99 (commented):
In one case when I went back via the drawer back to site-features and re-selected jetpack, it was added to the install queue but was immediately removed and was never installed, can you please check this as well?

arunshenoy99 (approved):

PR #128, opened 2022-12-11
Set no-title Page Template on wp_insert_post for Homepage

0aveRyan:
Notes:
* `no-title` maps to the Page Template available in the YITH Wonder theme.

Currently when Onboarding is injecting posts they're all getting the standard Template. This creates a disconnect between Live Previews we generate correctly in Onboarding, but how content renders in Page templates from the Block Theme.

I've aimed to leave this open to other future template declarations that I know we'll need to have capabilities for.

### DIFF
--- a/includes/RestApi/SitePagesController.php
+++ b/includes/RestApi/SitePagesController.php
@@ -94,7 +94,7 @@ class SitePagesController {
 			 \update_option( Options::get_option_name( 'show_on_front', false ), 'page' );
 		}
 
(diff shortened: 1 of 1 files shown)


arunshenoy99 (commented):
Pushed small changes to

1. Remove array which does not have `page_template` and use the `$post` variable.
2. Make function accept `$is_template_no_title` instead of checking titles.

@0aveRyan please confirm if these look good.

PR #131, opened 2022-12-12
Use white text for selected color palette text label

0aveRyan:
Use white text for selection state on color palettes.


arunshenoy99 (approved):

arunshenoy99 (commented):
Do we need white text for typography as well?

0aveRyan:
@arunshenoy99 yes, good catch. lets get this merged and make that a fast-follow too. it's more pronounced in colors because the titles are shorter

PR #132, opened 2022-12-12
Improve text copy in half-dozen sidebars

0aveRyan:
Polish text strings in sidebars, particularly before we get to localization discussions.

### DIFF
--- a/src/OnboardingSPA/pages/Steps/BasicInfo/Sidebar/LearnMore/contents.js
+++ b/src/OnboardingSPA/pages/Steps/BasicInfo/Sidebar/LearnMore/contents.js
@@ -10,7 +10,7 @@ const getContents = ( brandName ) => {
 			subheading: sprintf(
 				/* translators: %s: site */
 				__(
-					`Setup how your %s will present to the public, search engines and visitors’ web browsers.`,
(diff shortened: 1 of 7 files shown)


src/OnboardingSPA/pages/Steps/BasicInfo/Sidebar/LearnMore/contents.js
```diff
@@ -30,23 +30,13 @@ const getContents = ( brandName ) => {
 					description: sprintf(
 						/* translators: 1: Brand 2: Site */
 						__(
-							`Enter the best way for WordPress, web browsers and search engines to identify your %s %s by entering your Title, Description and Logo.`,
+							`Loading your site details, logo and social graph helps not just launch your site but have it found looking great in feeds.`,
```
arunshenoy99, line 33: @0aveRyan should all occurrences of `site` be replaced with `store` dynamically here? We use `store` for everything else in basic info.
Screenshot 2022-12-12 at 12 24 41 PM

0aveRyan: Yes, fine if we fix post-release though let's just note.

PR #135, opened 2022-12-12
No Site Title: Hashed Child Theme slug

arunshenoy99:


0aveRyan (approved):
Yep, this is exactly what I was envisioning by hashing the site url. Unique per-site on account, overwhelmingly likely unique per-site on box and likely unique across farm.

This is only when folks don't name a site. Which is the edge case. I think we're well covered!

PR #138, opened 2022-12-20
Press 2 458 Loader for Radio label controls

officiallygod:


arunshenoy99 (commented):
Can we separate the radio control from the actual skeleton and make them two separate components? Let me know if you need some ideas there. We also need to think of separating `src/OnboardingSPA/components/LivePreview/LivePreviewSkeleton/index.js` from the live preview components itself but we can explore as part of another story. This is just to ensure that we are writing reusable code that will reduce tedious work long term. Rest everything looks good.

arunshenoy99 (changes requested):

arunshenoy99 (approved):

PR #146, opened 2023-01-05
Basic Info FadeIn State

Yashita101:


src/OnboardingSPA/pages/Steps/BasicInfo/stylesheet.scss
```diff
@@ -3,6 +3,12 @@
     align-items:center;
     flex-direction: column;
     justify-content: center;
+    animation: fadeIn 500ms ease-in-out forwards;
```
officiallygod, line 6: Can we get rid of all this CSS as we won't be needing it right?

Yashita101: Yes, removing it

src/OnboardingSPA/pages/Steps/BasicInfo/stylesheet.scss
```diff
@@ -4,6 +4,11 @@
     flex-direction: column;
     justify-content: center;
 
+    &__loadState {
+        opacity: 0;
+        pointer-events: none;
+    }
```
officiallygod, line 10: Do we need this `__loadState`  CSS?

officiallygod (approved):

officiallygod:
Let's wait for PRESS 2-462 to be merged before we merge this Story, as this has changes from that story.

arunshenoy99:
Investigate if we can wait on the site logo to load as well instead of waiting only on flow and social API's

arunshenoy99 (changes requested):
Pending: Investigate if we can wait on the site logo to load as well instead of waiting only on flow and social API's

src/OnboardingSPA/pages/Steps/BasicInfo/basicInfoForm.js
```diff
@@ -129,94 +129,106 @@ const BasicInfoForm = () => {
...
-							content.siteTitle.hint,
-							'wp-module-onboarding'
-						) }
-						placeholder={ sprintf(
-							__(
-								content.siteTitle.placeholder,
+		<Animate
+			type={"fade-in"}
+			after={
+				typeof flowData === 'object' && typeof socialData === 'object'
```
officiallygod, line 135: Can we use Double Quotes as I saw in the docs usage.
Screenshot 2023-01-16 at 3 41 20 PM

officiallygod (commented):
Also in basicInfoForm.js 
function setDefaultData() {
		if ( isLoaded ) {
			setSiteLogo( flowData?.data.siteLogo );
			setSiteTitle( flowData?.data.blogName );
			setSiteDesc( flowData?.data.blogDescription );
		}
	}
	
	Can we add a `?` after **data** too for all three cases? In case flow API fails??

Yashita101:
> Pending: Investigate if we can wait on the site logo to load as well instead of waiting only on flow and social API's

The rendering of the image of the user depends on their internet speed, which is not possible to track via our code, as we can verify if the url is valid and not whether the rendering is complete. Hence this condition for siteLogo cannot be checked.

arunshenoy99:
> > Pending: Investigate if we can wait on the site logo to load as well instead of waiting only on flow and social API's
> 
> The rendering of the image of the user depends on their internet speed, which is not possible to track via our code, as we can verify if the url is valid and not whether the rendering is complete. Hence this condition for siteLogo cannot be checked.

Have some idea around tracking, will create a new story.

arunshenoy99 (changes requested):
@Yashita101 I just merged `trunk` into this branch to get a good idea of the changes and noticed something: can we please remove all the changes in `src/OnboardingSPA/pages/Steps/Ecommerce/StepAddress/index.js`? This change deletes `Animate` from `StepAddress` and we definitely need `Animate` in our address step.

arunshenoy99 (approved):
Looks good, will create a story with details to handle the site logo render.

PR #153, opened 2023-01-16
Fallback to customer data if flow param/option does not exist

arunshenoy99:


includes/ModuleController.php
```diff
@@ -10,89 +10,88 @@
 
 /**
  * Class ModuleController
- *
```
arunshenoy99, line 13: Changes in `includes/ModuleController.php` other than the formatting are
- Switch from `date` to `gmdate`, phpcs found a code standards violation `date() is affected by runtime timezone changes which can cause date/time to be| incorrectly displayed. Use gmdate() instead.`
- Use the `is_ecommerce_plan` function

officiallygod (approved):

PR #155, opened 2023-01-23
Modify DesignStateHandler to allow switching back to Wonder via a confirmation dialog

arunshenoy99:
1. Allows switching back via dialog when user re-visits onboarding and does not have Wonder active.
2. Allows switching back via dialog when user re-visits the design steps via drawer/header-navigation after generating the child theme.
3. Prevents unnecessary API calls in the error state.


officiallygod (approved):

avneet-raj:
https://paste.pics/86f6b65748bbcf180403c394e9a815c6

Hi @arunshenoy99 , according to the description we have options for YES and Exit to Wordpress. But here it is continue.

avneet-raj:
https://user-images.githubusercontent.com/79664985/215778591-211ae884-99af-4faa-aafb-7cc6be2f76a2.mp4

when we change the theme from admin dashboard and then try to change the theme style without refreshing the page, it gives something like this.
https://paste.pics/d5d5a922dd7f4658e11d95579405ab9a

PR #157, opened 2023-01-24
PRESS 2 498 | Audit memoization candidates

officiallygod:
Added a few improvements so as to reduce the number of rerenders in the commonly used pages throughout the Onboarding Module


Yashita101 (approved):

src/OnboardingSPA/components/LivePreview/BlockPreview/index.js
```diff
@@ -76,10 +76,9 @@ const BlockPreview = ( {
 		}
 	}, [ storedPreviewSettings ] );
 
-	return (
-		<div className={ `live-preview__container-${ styling }` }>
-			{ loading && (
-				<div className="live-preview__container--is-skeleton">
+	const SkeletonLivePreview = memo(() => {
+		return (
+			<div className="live-preview__container--is-skeleton">
```
Yashita101, line 81: Hey
Everything looks great, let's just resolve the conflict here, TIA.

Yashita101 (changes requested):

src/OnboardingSPA/components/LivePreview/BlockPreview/index.js
officiallygod: Isn't this fixed already?

Yashita101 (approved):

arunshenoy99 (changes requested):
Hey Allen, added one code comment and then these are a few things we need to take a look at 
**Top 3 Potential Memoization Opportunities** (I am not an expert so might be wrong on a few of these)

1. The Common Layout used across onboarding is being re-rendered causing a domino effect re-rendering all its children as well?
Screenshot 2023-02-01 at 1 36 45 PM

2. Sidebars are rendered 3 times on every page + 1 render on open.
Screenshot 2023-02-01 at 12 29 45 PM

3. The drawer is being rendered multiple times?
Screenshot 2023-02-01 at 1 51 29 PM

**Important Note**
Can we check if the changes here are the correct way to use memo? If I am not wrong effective memoization flows starting from the parent to the child so that a small parent re-render does not result into a big expensive child re-render(like our previews and the common layout), please take a look at a few of these links that helped me understand memoization, https://dmitripavlutin.com/use-react-memo-wisely/, https://medium.com/@vitaliysteffensen/how-i-eliminate-all-unnecessary-rerenders-in-react-79505deeedea and https://github.com/WordPress/gutenberg/issues/25780

Based on the note I am not sure if we have missed more potential opportunities, please do take a look at those other than the top 3 as well.

arunshenoy99 (commented):
Let's remove the build changes? they can cause a conflict anything changes in trunk

PR #167, opened 2023-02-05
Added JS and SASS Lint Checks

officiallygod:
1.  Checks for JS and SASS coding standards issues.
2. Prevents usage of restricted packages/imports via the ESLint config.


officiallygod:
LGTM!

PR #168, opened 2023-02-06
use branding variables in image uploader loading state gradient

Yashita101:


src/OnboardingSPA/styles/_branding.scss
```diff
@@ -27,6 +27,8 @@ body {
 		--nfd-onboarding-border: 219, 219, 219;
 		--nfd-onboarding-drawer-icon-fill: var(--nfd-onboarding-tertiary);
 		--nfd-onboarding-drawer-icon-active-fill: var(--nfd-onboarding-light);
+		--nfd-onboarding-loader-start: var(--nfd-onboarding-primary);
+		--nfd-onboarding-loader-end: var(--nfd-onboarding-tertiary);
```
arunshenoy99, line 31: Should we be adding these to the Newfold CSS as well?

Yashita101: I thought so too, but I do not see `--nfd-onboarding-tertiary`  and some other corresponding branding variables in `_interface.css` . We would need to get the respective data.

arunshenoy99: Sounds good looks like we'll be doing that later on, rest all looks good to me. Approving, will do a QA and merge this.

arunshenoy99 (approved):

PR #169, opened 2023-02-06
settings api - tiktok and yelp url fixes

abhijitb:


includes/RestApi/SettingsController.php
```diff
@@ -206,13 +208,30 @@ public function get_current_settings() {
...
 			$social_data['twitter_site'] = 'https://www.twitter.com/' . $twitter_handle;
 		}
 
+		// handle other social urls for onboarding
+		foreach ( $social_data['other_social_urls'] as $index => $social_url ) {
+			// remove the indexed url from array
+			unset( $social_data['other_social_urls'][ $index ] );
+
+			switch ( $social_url ) {
+				case ( -1 < stripos( $social_url, 'yelp.com' ) ):
```
arunshenoy99, line 223: Let's use that regex we discussed about?

arunshenoy99 (changes requested):
Hey @abhijitb did a quick test before reviewing all of the code, I seem to always be in either of these 2 conditions both of which do not clear the yelp/tiktok URL's.
1. A `mastodon_url` has appeared causing the API to fail (only when Yoast is active) ss-1
2. Request:
ss-2
Response:
ss-3

includes/RestApi/SettingsController.php
abhijitb: regex check added

arunshenoy99 (approved):

PR #178, opened 2023-02-22
Checks to prevent re-queuing plugins

officiallygod:


includes/RestApi/PluginsController.php
```diff
@@ -322,6 +322,17 @@ public function set_site_features( \WP_REST_Request $request ) {
...
+				// If the Plugin exists and is activated
+				if ( PluginInstaller::exists( $plugin, $decision ) ) {
+					continue;
+				}
+
+				// If the Plugin is already in queue
+				$position_in_queue = PluginInstallTaskManager::status( $plugin );
+				if ( false !== $position_in_queue ) {
+					continue;
+				}
```
arunshenoy99, line 334: We don't need to do this, the `add_to_queue` function takes care of preventing duplication in the queue.

arunshenoy99 (changes requested):

includes/RestApi/PluginsController.php
officiallygod: Yeah Sure, Done.

arunshenoy99 (approved):

PR #179, opened 2023-02-22
Add/press2 383 flow data update on version change

Yashita101:


includes/Data/Flows.php
```diff
@@ -13,6 +13,8 @@ final class Flows {
 	 * @var array
 	 */
 	protected static $data = array(
+		'version'              => '0.1',
```
arunshenoy99, line 16:
```suggestion
		'version'              => '1.0.0',
```
Let's start following semantic versioning for this as well?

includes/Services/FlowService.php
```diff
@@ -21,8 +21,10 @@ public static function initalize_flow_data() {
 		if ( ! ( $flow_data ) ) {
 			return \update_option( Options::get_option_name( 'flow' ), $default_flow_data );
 		}
-		$updated_flow_data = self::update_flow_data_recursive( $default_flow_data, $flow_data );
-		return \update_option( Options::get_option_name( 'flow' ), $updated_flow_data );
+		if ( ! array_key_exists( 'version', $flow_data ) || strcmp( $flow_data['version'], $default_flow_data['version'] ) !== 0 ) {
```
arunshenoy99, line 24: Since we know the exact level of nesting at which the `version` we are looking for will exist, let's use `isset` instead of `array_key_exists` which is a bit faster?

includes/Services/FlowService.php
```diff
@@ -68,6 +70,12 @@ private static function update_flow_data_recursive( $default_flow_data, $flow_da
 		$exception_list    = Flows::get_exception_list();
 		$updated_flow_data = array();
 		foreach ( $default_flow_data as $key => $value ) {
+			// To update the options with the recent version of flow data
+			if ( strcmp( 'version', $key ) === 0 ) {
+				$updated_flow_data[ $key ] = $value;
+				continue;
+			}
```
arunshenoy99, line 77: Small question here, since this a recursive function, let's say we add a version to another key in flow data, will that version also get updated with the value from `default_flow_data`?

arunshenoy99 (changes requested):
Looks good, suggesting a couple of minor changes and a question.

arunshenoy99 (approved):

PR #183, opened 2023-02-23
Improve plugin status detection

arunshenoy99:
When plugin installation is in progress via the WP_Cron and the plugin directory is in a half-baked state, Onboarding is unable to determine if the plugin has been installed or not, causing it to default to a not installed state. This PR checks if a plugin is being installed and then waits for the installation task to complete before defaulting to the not installed state.

### DIFF
--- a/includes/RestApi/PluginsController.php
+++ b/includes/RestApi/PluginsController.php
@@ -1,6 +1,7 @@
 <?php
 namespace NewfoldLabs\WP\Module\Onboarding\RestApi;
 
+use NewfoldLabs\WP\Module\Onboarding\Data\Options;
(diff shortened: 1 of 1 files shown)


Yashita101 (approved):

avneet-raj:
Tested on my local for these 3 scenarios:
1. When we go very fast ecom step. Sometimes the woocommerce plugin gets installed and activated within time i.e 30sec but sometimes it gets only installed and not activated and then we refresh the page the plugin gets activated and then we are good to go.
2. When we go with normal approach with selecting every elements to the ecom steps, after 30 sec wait the plugin gets activated and we can move ahead.
3. When the woocommerce plugin is already there and activated, then there is no wait for 30 sec for the ecom step to load.

PR #185, opened 2023-03-01
PRESS 2 111 | Fix Social Data to be handled outside Flow

officiallygod:


.github/workflows/lint-check-spa.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
       # Installs @wordpress/scripts for lint checks if it does not exist in the cache.
       - name: Install dependencies
-        run: npm i @wordpress/scripts
+        run: npm i @wordpress/scripts lodash
```
arunshenoy99, line 44: We can remove this and add `// eslint-disable-next-line import/no-extraneous-dependencies` for now wherever necessary, this is causing caching issues

officiallygod: Sure

Yashita101 (approved):

PR #189, opened 2023-03-03
Add Crazy Domains Data and Branding

arunshenoy99:

PR #190, opened 2023-03-06
Add dynamic brand configuration to hide sidebar illustrations for crazy domains

arunshenoy99:
https://github.com/newfold-labs/wp-module-onboarding/pull/189 needs to be merged before this.


arunshenoy99:
> Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> 
> Screenshot 2023-03-06 at 7 47 41 PM



> Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> 
> Screenshot 2023-03-06 at 7 47 41 PM

@officiallygod Good question, I thought of this initially, it makes sense to add the `brandConfig` check to `IllustrationPanel` just for this case where we are adding the Panel based on the config, but considering broader cases like conditionally determining the order and kind of components to be shown (or) changing the behavior of the sidebar as defined by the config which might arise eventually, it makes more sense to leave this logic in the step sidebar components. As an example, let's say Brand 1 does not require an `IllustrationPanel` that shows the image but might require another component or set of components in its place (or) Brand 2 simply needs a different structure that does not have any of the components used above rather has its own set of Components as defined in the `brandConfig` and many more cases like these are examples where taking this approach (even though tedious), saves us a lot of time long term :)

officiallygod:
> > Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> > Screenshot 2023-03-06 at 7 47 41 PM
> 
> > Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> > Screenshot 2023-03-06 at 7 47 41 PM
> 
> @officiallygod Good question, I thought of this initially, it makes sense to add the `brandConfig` check to `IllustrationPanel` just for this case where we are adding the Panel based on the config, but considering broader cases like conditionally determining the order and kind of components to be shown (or) changing the behavior of the sidebar as defined by the config which might arise eventually, it makes more sense to leave this logic in the step sidebar components. As an example, let's say Brand 1 does not require an `IllustrationPanel` that shows the image but might require another component or set of components in its place (or) Brand 2 simply needs a different structure that does not have any of the components used above rather has its own set of Components as defined in the `brandConfig` and many more cases like these are examples where taking this approach (even though tedious), saves us a lot of time long term :)

@arunshenoy99 I did think that but then I also thought that it won't be that good if we are sending brand-specific logic per step ( i.e around 15 steps ) of logic for conditional rendering from the backend making even the store very heavy.
But Sureee else everything looks goood.

officiallygod:
> > Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> > Screenshot 2023-03-06 at 7 47 41 PM
> 
> > Why don't we put `IllustrationPanel` specific logic in one place inside the same file? Something like this?? As the config passed is not step-specific and instead **brand** specific.
> > Screenshot 2023-03-06 at 7 47 41 PM
> 
> @officiallygod Good question, I thought of this initially, it makes sense to add the `brandConfig` check to `IllustrationPanel` just for this case where we are adding the Panel based on the config, but considering broader cases like conditionally determining the order and kind of components to be shown (or) changing the behavior of the sidebar as defined by the config which might arise eventually, it makes more sense to leave this logic in the step sidebar components. As an example, let's say Brand 1 does not require an `IllustrationPanel` that shows the image but might require another component or set of components in its place (or) Brand 2 simply needs a different structure that does not have any of the components used above rather has its own set of Components as defined in the `brandConfig` and many more cases like these are examples where taking this approach (even though tedious), saves us a lot of time long term :)

@arunshenoy99 I did think that but then I also thought that it won't be that good if we are sending brand-specific logic per step ( i.e around 15 steps ) of logic for conditional rendering from the backend making even the store very heavy.
But sure else everything looks good.

PR #195, opened 2023-03-16
Razorpay switch to plugin worker API

Yashita101:


includes/Data/Plugins.php
```diff
@@ -102,6 +98,11 @@ final class Plugins {
 			'url'      => 'https://hiive.cloud/workers/plugin-downloads/yith-woocommerce-wishlist',
 			'path'     => 'yith-woocommerce-wishlist-extended/init.php',
 		),
+		'nfd_slug_woo_razorpay'             => array(
+			'approved' => true,
+			'url'      => 'https://hiive.cloud/workers/plugin-downloads/razorpay',
+			'path'     => 'woo-razorpay/woo-razorpay.php',
+		),
```
arunshenoy99, line 105: @aulisius @ashrithapradeep FYI

aulisius: Does this the change API we need to use to queue plugin installs?

arunshenoy99: Yup, I think have to change the commerce module currently using the API with woo-razorpay to use nfd_slug_woo_razorpay.

arunshenoy99 (approved):

PR #203, opened 2023-03-28
General Onboarding : Add brand specific enabled_flows configuration

arunshenoy99:
1. Adds an `enabled_flows` configuration brand wise to give us better control over the flows enabled on a particular brand.
2. Enables `wp-setup` (General) flow for Bluehost and Bluehost India


includes/ModuleController.php
```diff
@@ -59,43 +62,117 @@ public static function module_switcher() {
...
+		$brand_enabled_flows = Flows::get_flows();
 
-		if ( isset( $_GET['flow'] ) && 'ecommerce' === \sanitize_text_field( $_GET['flow'] ) ) {
-			return true;
+		foreach ( $brand_enabled_flows as $flow => $enabled ) {
+			if ( ! $enabled ) {
+				continue;
+			}
+
+			switch ( $flow ) {
```
officiallygod, line 78: If we do it like this? cause even if the `flow=wp-setup` as `case 'ecommerce'` ( woocommerce is active ) runs first we will get true there and actually `case 'wp-setup'` should have been the one that should return true but it would never even run so logically looking it would be weird though it will work.
Should it not be like if the flow is **wp-setup** we don't even need to check woocommerce or cust data? just check the sign-up date and go forward? why go through the extra overhead?

arunshenoy99: Nope, that would mean anyone can manipulate the flow by just setting the flow param. I've written it precisely in the way you describe and it is not a mistake, anyone eligible for commerce Onboarding should see it irrespective of the flow param.

I think we had some issues activating the module because of customer data not being set by the time the redirect happens which is why we skip all checks if `flow=ecommerce` is present, we're beyond such issues now with the new "isEcommerce/Woocommerce" and "plugin_install_date" fallbacks and will probably remove that logic after some thorough testing (so that someone not eligible for commerce Onboarding does not get access to it by just passing `flow=ecommerce`, it is not that a critical thing which is why I have not bought this up yet). 

The scenario of passing `flow=wp-setup` for a commerce customer (woo active) that you describe will arise in 2 cases
1. Someone manipulates the URL's flow param to access the wp-setup flow.
2. We are setting the wrong value on the flow param during the redirect.

Either way, the code written in that manner will always ensure that users always get the correct Onboarding flow that they are eligible for on that brand irrespective of the flow param.

arunshenoy99: Also `enabled_flows` is designed to have keys in the order of priority, which means that a key defined earlier in the associative array gets higher priority over keys below it and the last enabled key will act as the fallback if none of the other flows criteria is met.

Yashita101 (approved):

includes/ModuleController.php
Yashita101: The current approach looks good then. Thanks.

arunshenoy99: @officiallygod, to add some context for when you're back, we will not be taking the approach **yet** but rather rely on the flow param until we know that the fallback on Woocommerce/Customer data from hosting is reliable. So we should be fine as long as we are setting the right flow param value.

PR #217, opened 2023-04-19
Text and urls added for the default brand

Yashita101:


includes/Data/Brands.php
```diff
@@ -14,6 +14,62 @@ final class Brands {
 	public static function get_brands() {
 
 		return array(
+			'newfold'       => array(
+				'name'                        => 'your web host',
```
arunshenoy99, line 18: Should we be passing `your web host` to the `__` helper function, since it is not a proper noun?

arunshenoy99 (changes requested):
Should we resolve merge conflicts in the base branch and then propagate the latest code to this branch?

includes/Data/Data.php
```diff
@@ -35,10 +35,11 @@ public static function runtime() {
 	 */
 	public static function current_brand() {
 		$brands = Brands::get_brands();
+		// $default_brand = Brands::get_default_brand();
 
 		return array_key_exists( NFD_ONBOARDING_PLUGIN_BRAND, $brands ) ?
 			$brands[ NFD_ONBOARDING_PLUGIN_BRAND ] :
-			array( 'brand' => NFD_ONBOARDING_PLUGIN_BRAND );
+			array_replace( $brands['bluehost'], Brands::get_default_brand() );
```
arunshenoy99, line 42: This logic should be in the get_default_brand function.

includes/Data/Data.php
```diff
@@ -35,10 +35,11 @@ public static function runtime() {
 	 */
 	public static function current_brand() {
 		$brands = Brands::get_brands();
+		// $default_brand = Brands::get_default_brand();
```
arunshenoy99, line 38:
```suggestion
```
Can we remove the commented code?

arunshenoy99 (approved):

PR #228, opened 2023-05-01
Enhance preview step to auto advance on custom design checkbox selection

arunshenoy99:
Auto progresses to colors step when colors and typography checkbox is ticked manually by a user. Maintains history so that the user can go back to the larger preview via the browser back button or our header/drawer navigation.


officiallygod (approved):
Looks Good to me.

PR #229, opened 2023-05-02
PRESS-2 761 - Implement Interface Skeleton

officiallygod:


src/OnboardingSPA/components/NewfoldInterfaceSkeleton/index.js
```diff
@@ -0,0 +1,163 @@
...
+			{ ...navigateRegionsProps }
+			ref={ useMergeRefs( [ ref, navigateRegionsProps.ref ] ) }
+			className={ classnames(
+				className,
+				'nfd-interface-interface-skeleton',
+				navigateRegionsProps.className,
+				!! footer && 'has-footer'
+			) }
+		>
+			{ !! drawer && (
```
arunshenoy99, line 82: Can we just check why `!!` was used?

arunshenoy99 (commented):
Did not do an in-depth check since this code is from Gutenberg, suggested a minor change that we might have missed when pasting. Rest all looks good.

src/OnboardingSPA/components/NewfoldInterfaceSkeleton/index.js
officiallygod: The double exclamation point, or double bang, converts a truthy or falsy value to “true” or “false”. In other words, it operates exactly like Boolean(value) would. [Ref](https://medium.com/@chirag.viradiya_30404/whats-the-double-exclamation-sign-for-in-javascript-d93ed5ad8491#:~:text=The%20double%20exclamation%20point%2C%20or,like%20Boolean(value)%20would.)

arunshenoy99: Thanks

arunshenoy99 (approved):

PR #230, opened 2023-05-02
Enhance/make flow data robust

arunshenoy99:
@Yashita101 please go ahead and merge this PR once lint passes, LGTM.


officiallygod:
The option **flow_preset** is that set by us?
Screenshot 2023-05-11 at 6 13 13 PM

includes/Services/FlowService.php
```diff
@@ -0,0 +1,262 @@
...
+	public static function initalize_flow_data() {
+		$default_flow_data = self::get_default_flow_data();
+		$flow_data         = self::read_flow_data_from_wp_option();
+
+		if ( ! ( $flow_data ) ) {
+			return \update_option( Options::get_option_name( 'flow' ), $default_flow_data );
+		}
+
+		if ( ! isset( $flow_data['version'] ) || strcmp( $flow_data['version'], $default_flow_data['version'] ) !== 0 ) {
+			$flow_data_upgrades = NFD_ONBOARDING_DIR . '\\includes\\Data\\Flows\\Upgrades\\' . $default_flow_data['version'] . '.php';
```
officiallygod, line 27: Hey, do you know if this code is windows specific? I think this won't work on Mac and Linux, and as the site will be running on a Linux env, maybe we can have a more generalized way ?
Screenshot 2023-05-11 at 7 04 41 PM

officiallygod (changes requested):
@Yashita101, I had a doubt, suppose say the user is on v1.0.0 and he goes through onboarding, then he goes through onboarding again in v3.0.0 and doesn't go through v2.0.0.

And say that the update on v3.0.0 is dependent on v2.0.0, like 

v1.0.0 `colorStyle`

v2.0.0 `colorStyle -> colorPalette`

v3.0.0 `colorPalette -> colorOptions`

So it will search for **colorPalette** in v3 but won't get that key as that is **colorStyle**

includes/Services/FlowService.php
```diff
@@ -0,0 +1,262 @@
...
+			if ( $flow_data_upgrades ) {
+				require $flow_data_upgrades;
```
officiallygod, line 29: Should we user require or include? cause if require fails it returns a fatal error and the application might crash.
https://www.w3schools.com/php/php_includes.asp#:~:text=The%20include%20and,script%20will%20continue

includes/RestApi/Themes/ThemeGeneratorController.php
```diff
@@ -446,12 +446,6 @@ private function validate_and_sanitize_flow_data( $flow_data ) {
 			$flow_data['theme']['template'] = Themes::get_flow_default_theme_slug( $current_flow );
 		}
 
-		if ( $flow_data['customDesign'] ) {
```
officiallygod, line 449: Since we are not checking this should we have an alternative check with the new keys here?

src/OnboardingSPA/pages/Steps/BasicInfo/basicInfoForm.js
```diff
@@ -24,7 +24,7 @@ const BasicInfoForm = () => {
 
 	const [ siteTitle, setSiteTitle ] = useState( '' );
 	const [ siteDesc, setSiteDesc ] = useState( '' );
-	const [ siteLogo, setSiteLogo ] = useState( 0 );
+	const [ siteLogo, setSiteLogo ] = useState( { id: 0, url: '' } );
```
officiallygod, line 27: If we are setting this data pattern in flows, why are we again setting it here? should we not be using the one from the backend? else if we decide to make `id 1` then we need to change the backend and frontend?

includes/RestApi/Themes/ThemeGeneratorController.php
Yashita101: It is not required, as we are getting the color and font slug data from global styles.

includes/Services/FlowService.php
Yashita101: Yes agreed, considering that while writing to the updated version files, we can avoid such fatal errors, hence included `require`. However, to avoid such scenarios where the site might crash, shall revert it to `includes`

includes/Data/Flows/Flows.php
```diff
@@ -147,6 +126,50 @@ final class Flows {
...
+	 *
+	 * - the value of old_key i.e OldKey is the key name in the option that has been modified to NewKey in the blueprint
+	 * - the value of new_key i.e NewKey is the key name that is be replaced in the option
+	 * - OldKey should also be manually updated to the NewKey in the blueprint
+	 * - retain_existing_value (Boolean): true if the value in the option for the OldKey is to be retained,
+	 *   false if the NewKey is to have the default value that is set in the blueprint and discard the value set in the option for OldKey
+	 *
+	 * @var array
+	 */
+	protected static $fixes = array(
```
officiallygod, line 140: Just wanted to ask from now how are we migrating changes? like what is the use-case of `fixes` vs `upgrades`?

includes/RestApi/FlowController.php
```diff
@@ -181,7 +161,7 @@ public function save_onboarding_flow_data( \WP_REST_Request $request ) {
 		}
 
 		// save data to database
-		if ( ! $this->update_wp_options_data_in_database( $flow_data ) ) {
+		if ( ! \update_option( Options::get_option_name( 'flow' ), $flow_data ) ) {
```
officiallygod, line 164: Why are we not using a method for this from FlowService? considering we have one for `FlowService::read_flow_data_from_wp_option();`

Cause I feel I might miss something while doing copy-paste.

includes/RestApi/FlowController.php
```diff
@@ -106,24 +97,15 @@ public function save_onboarding_flow_data( \WP_REST_Request $request ) {
...
-		$flow_data = $this->read_details_from_wp_options();
-		if ( ! $flow_data ) {
-			$flow_data              = Flows::get_data();
-			$flow_data['createdAt'] = time();
-			// update default data if flow type is ecommerce
-			$flow_data = $this->update_default_data_for_ecommerce( $flow_data );
-			$this->save_details_to_wp_options( $flow_data );
+		$default_flow_data = FlowService::get_default_flow_data();
+		$mismatch_key      = FlowService::find_mismatch_key( $params, $default_flow_data );
+		if ( \is_wp_error( $mismatch_key ) ) {
```
officiallygod, line 102: Just a doubt, do we need the `\` for `is_wp_error`?

includes/Data/Flows/Upgrades/1.0.0.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+namespace NewfoldLabs\WP\Module\Onboarding\Data\Flows\Upgrades;
+
+use NewfoldLabs\WP\Module\Onboarding\Data\Options;
+
+if ( \get_option( Options::get_option_name( 'flow' ), false ) ) {
+	$flow_data['data']['colorStyle'] = ! empty( $flow_data['data']['palette']['slug'] ) ? $flow_data['data']['palette']['slug'] : '';
+	$flow_data['data']['fontStyle']  = ! empty( $flow_data['data']['typography']['slug'] ) ? $flow_data['data']['typography']['slug'] : '';
+	\update_option( Options::get_option_name( 'flow', false ), $flow_data );
```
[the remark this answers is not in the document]
Yashita101: you're right, missed on that, thanks

includes/Services/FlowService.php
```diff
@@ -0,0 +1,262 @@
...
+
+		if ( ! isset( $flow_data['version'] ) || strcmp( $flow_data['version'], $default_flow_data['version'] ) !== 0 ) {
+			if ( isset( $flow_data['version'] ) ) {
+				UpgradeHandler::maybe_upgrade( $flow_data['version'], $default_flow_data['version'] );
+			}
+			$updated_flow_data = self::update_flow_data_recursive( $default_flow_data, $flow_data );
+			// To update the options with the recent version of flow data
+			$updated_flow_data['version'] = $default_flow_data['version'];
+			return \update_option( Options::get_option_name( 'flow' ), $updated_flow_data );
+		}
```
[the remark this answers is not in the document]
Yashita101: returning false if no if conditions hold true

includes/Services/FlowService.php
```diff
@@ -0,0 +1,262 @@
...
+	public static function initalize_flow_data() {
+		$default_flow_data = self::get_default_flow_data();
+		$flow_data         = self::read_flow_data_from_wp_option();
+
+		if ( ! ( $flow_data ) ) {
+			return \update_option( Options::get_option_name( 'flow' ), $default_flow_data );
+		}
+
+		if ( ! isset( $flow_data['version'] ) || strcmp( $flow_data['version'], $default_flow_data['version'] ) !== 0 ) {
+			if ( isset( $flow_data['version'] ) ) {
```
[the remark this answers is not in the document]
Yashita101: agreed, have removed the if condition altogether to allow the routines to run for existing customers. For the new customers, the first if condition would be satisfied anyway.

officiallygod (approved):

includes/Services/FlowService.php
```diff
@@ -0,0 +1,271 @@
...
+					'Wrong Parameter Type Provided : ' . $key . ' => Indexed Array. Expected: Associative Array',
+					array( 'status' => 400 )
+				);
+			}
+
+			// If the Database value is Empty/Indexed Array, to avoid Associative arrays to be overwritten (Eg: data)
+			if ( self::is_array_indexed( $value ) || self::is_array_indexed( $default_flow_data[ $key ] ) ) {
+				$flow_data[ $key ] = $params[ $key ];
+				continue;
+			}
```
arunshenoy99, line 187: We can skip this, considering the only thing we need to validate is `params`?

includes/Services/FlowService.php
```diff
@@ -191,42 +385,19 @@ private static function update_data_for_ecommerce( $data ) {
 		if ( 'ecommerce' === $flow_type ) {
 			// update default data with ecommerce data
 			$data['data']['topPriority']['priority1'] = 'selling';
+			$data['data']['siteType']['referTo']      = 'business';
```
arunshenoy99, line 388:
```suggestion
```
We do not use this after the site classification API went live, good to remove.

includes/Services/FlowService.php
```diff
@@ -23,9 +70,7 @@ class FlowService {
 	public static function get_flow_data() {
```
arunshenoy99, line 70: Should we be renaming this to `get_data` as well to stay consistent with the other ones?

includes/Services/FlowService.php
```diff
@@ -46,25 +91,23 @@ public static function update_flow_data( $params ) {
 			);
 		}
 
-		$flow_data = self::get_flow_data();
+		$default_data = self::get_default_data();
+		$mismatch_key = self::find_mismatch_key( $params, $default_data );
```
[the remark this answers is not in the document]
Yashita101: since we do not iterate through `$params` in the recursive call, a check is added using the `array_diff` of the keys at the level in the array and an error is thrown accordingly. Have removed the mismatch_key logic, thanks.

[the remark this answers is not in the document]
arunshenoy99: Do we need to explicitly check `$params` was my initial question, seems like irrespective of what the keys are in `$params` we take the keys from `$default_data` and take the value from `$params[$key]` after an `isset` check?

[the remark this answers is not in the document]
Yashita101: The explicit check is required if we need an error to be thrown gracefully for the wrong key entered by the user.
Otherwise, without that, there would not be an error and the keys would be retained as per the blueprint.

includes/Services/FlowService.php
```diff
@@ -91,37 +133,152 @@ public static function update_flow_data( $params ) {
...
+				\update_option( Options::get_option_name( 'site_icon', false ), $data['data']['siteLogo']['id'] );
+				\update_option( Options::get_option_name( 'site_logo', false ), $data['data']['siteLogo']['id'] );
 		} else {
 			\update_option( Options::get_option_name( 'site_icon', false ), 0 );
 			\delete_option( Options::get_option_name( 'site_logo', false ) );
 		}
 
-		if ( ! self::update_data_in_wp_option( $flow_data ) ) {
+		// Add the version key to the $data before updating to options data.
+		$updated_data = array_merge( array( 'version' => $options_data['version'] ), $data );
```
arunshenoy99, line 157:
```suggestion
	 $data['version'] = $options_data['version']
```

PR #232, opened 2023-05-04
Fix user capability check during login redirects

earnjam:
This should resolve the user capability check issue that was introduced with #231

### DIFF
--- a/includes/LoginRedirect.php
+++ b/includes/LoginRedirect.php
@@ -2,9 +2,7 @@
 namespace NewfoldLabs\WP\Module\Onboarding;
 
 use DateTime;
-use NewfoldLabs\WP\Module\Onboarding\Permissions;
 use NewfoldLabs\WP\Module\Onboarding\Data\Options;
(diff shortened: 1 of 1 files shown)


earnjam:
Just to add a side note here regarding removing the user capability check abstractions in this method: 

With any abstractions, it's a bit subjective as to where the line is to create them. But I think in general it is best to try to stay closer to core WP functions when possible. That may mean we have some minor redundancy or repetition of code when checking values (Example: checking that `$user` is an instance of `WP_User` here, or explicitly passing the `manage_option` capability), but it also means that anyone familiar with WordPress who is reading the code can easily see and understand what it's doing. It reduces some of the required mental overhead and file/class jumping when trying to understand or review a piece of code. 

With JS, every character counts toward the downloaded file size for the user, so making code shorter tends to be better. But with PHP I tend to lean toward being more verbose for clarity over slimming down into minor abstractions because there is no real performance cost for doing so.

PR #234, opened 2023-05-04
Split out redirect filters into separate methods

earnjam:
The `login_redirect` filter and the `newfold_sso_success_url` filters accept differing numbers of parameters. They also require different ways of determining the current user. The former does not have `$current_user` set and the latter does not have a `$user` global variable.

To make them both work correctly, and avoid a lot of logic to check values and types of multiple variables, I've created separate methods for each hook that each establish the logged in user, and then pass the url and user on to a single `filter_redirect` method for evaluation.


circlecube (approved):
I like using unique callbacks for these scenarios. 👍

PR #236, opened 2023-05-09
refactored hard coded logic from homepage menu

Yashita101:


includes/Data/Patterns.php
```diff
@@ -304,10 +289,58 @@ public static function get_theme_step_patterns_from_step( $step, $squash = false
...
+	public static function get_patterns_for_homepage_menu_slugs() {
+		// fetch the selected header menu slug from DB
+		$flow_data        = \get_option( Options::get_option_name( 'flow' ) );
+		$header_menu_slug = explode( '/', $flow_data['data']['partHeader'] )[1];
+		$header_slug      = ! empty( $header_menu_slug ) ? $header_menu_slug : 'site-header-left-logo-navigation-inline';
+
+		$homepage_slugs = array(
+			$header_slug,
+			'site-footer',
+		);
```
abhijitb, line 309: We dont need this array ... we simply save the header and footer slugs separately ... will make the code more readable .. right now we are using $homepage_slugs[0] lower down in the function

Yashita101: done, thanks

includes/Data/Patterns.php
```diff
@@ -281,8 +263,11 @@ public static function get_theme_step_patterns_from_step( $step, $squash = false
 		foreach ( array_keys( $pattern_slugs ) as $pattern_slug ) {
 			if ( true === $pattern_slugs[ $pattern_slug ]['active'] ) {
 				$pattern_name = $active_theme . '/' . $pattern_slug;
+				$pattern      = $block_patterns_registry->get_registered( $pattern_name );
```
arunshenoy99, line 266: This needs to go back inside the `if($block_patterns_registry->is_registered( $pattern_name ))` we cannot `get_registered` without first checking if it `is_registered`

includes/Data/Patterns.php
```diff
@@ -304,10 +289,54 @@ public static function get_theme_step_patterns_from_step( $step, $squash = false
...
+		$flow_data        = \get_option( Options::get_option_name( 'flow' ) );
+		$header_menu_slug = explode( '/', $flow_data['data']['partHeader'] )[1];
+		$header_slug      = ! empty( $header_menu_slug ) ? $header_menu_slug : 'site-header-left-logo-navigation-inline';
+		$footer_slug      = 'site-footer';
+
+		$homepage_style_slugs = array(
+			'homepage-1' => array(
+				'active'  => true,
+				'shown'   => true,
+				'content' => '',
```
arunshenoy99, line 311: What's the use of passing an empty `content` key here?

includes/Data/Patterns.php
```diff
@@ -304,10 +289,54 @@ public static function get_theme_step_patterns_from_step( $step, $squash = false
...
+	/**
+	 * Retrieve Homepage Menu Step Patterns
+	 *
+	 * @return array
+	 */
+	public static function get_patterns_for_homepage_menu_slugs() {
+		// fetch the selected header menu slug from DB
+		$flow_data        = \get_option( Options::get_option_name( 'flow' ) );
+		$header_menu_slug = explode( '/', $flow_data['data']['partHeader'] )[1];
+		$header_slug      = ! empty( $header_menu_slug ) ? $header_menu_slug : 'site-header-left-logo-navigation-inline';
```
arunshenoy99, line 304: Don't we have a function named `replace_header_menu_slug` that can handle this for us?
``` 
$flow_data  = \get_option( Options::get_option_name( 'flow' ) );
$header_menu_slug = explode( '/', $flow_data['data']['partHeader'] )[1];
``` 
Also, can we check if we can move duplicate logic from `get_theme_step_patterns_from_step` and this one into that function

includes/Data/Patterns.php
```diff
@@ -44,27 +44,7 @@ protected static function get_theme_step_patterns() {
...
-					),
-					'homepage-3'  => array(
-						'active' => true,
-						'shown'  => true,
-					),
-					'site-footer' => array(
-						'active' => true,
-					),
-				),
+				'homepage-styles' => self::get_patterns_for_homepage_menu_slugs(),
```
arunshenoy99, line 47: This needs to be a function reference that we can then call in `get_theme_step_patterns_from_step` else it seems like pattern parsing logic is running twice, once in this function call and once when the actual API request comes in and `get_theme_step_patterns_from_step` runs.

arunshenoy99: It's also running on every API request, even for steps other than homepage-styles

arunshenoy99 (changes requested):

src/OnboardingSPA/pages/Steps/DesignHomepageMenu/index.js
```diff
@@ -76,41 +53,32 @@ const StepDesignHomepageMenu = () => {
...
-			// update the header menu pattern if already selected
-			if (
-				currentData.data.partHeader ||
-				currentData.data.partHeader !== ''
-			) {
-				homepagePatterns[ 0 ] = currentData.data.partHeader;
+		Object.values( homepagePatternDataResp ).forEach(
+			( homepagePatternData ) => {
+				makeHomepagePattern.push( homepagePatternData.content );
 			}
```
arunshenoy99, line 59: Quick question, aren't lines 75-79 performing another `forEach`, is there anyway that we combine both of them? `setHomepagePattern` will be given an array of `contents` and `setHomepagePatternList` will be given an array of `slugs` both of which will be populated by a single `forEach` looping on `homepagePatternDataTemp?.body`

src/OnboardingSPA/pages/Steps/DesignHomepageMenu/index.js
```diff
@@ -76,41 +53,32 @@ const StepDesignHomepageMenu = () => {
...
 		}
 
+		setHomepagePattern( refactorPatterns( homepagePatternDataTemp?.body ) );
+
+		Object.values( homepagePatternDataTemp?.body ).forEach(
+			( homepagePatternData ) => {
+				homepagePatternList.push( homepagePatternData.slug );
+			}
+		);
+		setHomepagePatternList( homepagePatternList );
```
arunshenoy99, line 80: I see line 96 is also calling `setHomepagePatternList`, do we need to set the state twice? Changing states will cause re-renders, which is why we should try to minimize unnecessary state changes.

arunshenoy99 (changes requested):
This looks better, thanks @Yashita101, have suggested some more changes, please do take a look.

includes/Data/Patterns.php
```diff
@@ -213,16 +226,20 @@ public static function get_pattern_from_slug( $pattern_slug ) {
...
-			if ( isset( $slug_details['replace'] ) && true === $slug_details['replace'] ) {
-				unset( $patterns[ $slug ] );
-				$patterns = array_merge( array( $header_menu_slug => $slug_details ), $patterns );
+	private static function replace_header_menu_slug( $patterns ) {
+		// fetch the selected header menu slug from DB
+		$flow_data        = \get_option( Options::get_option_name( 'flow' ) );
+		$header_menu_slug = explode( '/', $flow_data['data']['partHeader'] )[1];
+		if ( ! empty( $header_menu_slug ) ) {
+			foreach ( $patterns as $slug => $slug_details ) {
+				if ( isset( $slug_details['replace'] ) && true === $slug_details['replace'] ) {
```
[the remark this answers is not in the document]
Yashita101: Yeah, since the header menu slug name has to be changed in the patterns list before the block grammar for each pattern is generated. Iterating through the list and checking for the `replace` key would not serve the purpose of changing the header menu slug.

includes/Data/Patterns.php
```diff
@@ -268,18 +279,15 @@ public static function get_theme_step_patterns_from_step( $step, $squash = false
...
-		}
 
 		foreach ( array_keys( $pattern_slugs ) as $pattern_slug ) {
 			if ( true === $pattern_slugs[ $pattern_slug ]['active'] ) {
+				if ( isset( $pattern_slugs[ $pattern_slug ]['replace'] ) && true === $pattern_slugs[ $pattern_slug ]['replace'] ) {
+					$pattern_slug_data = $pattern_slugs[ $pattern_slug ];
+					$header_menu_slug = self::replace_header_menu_slug();
+					$pattern_slug = ( ! empty( $header_menu_slug ) ) ? $header_menu_slug : $pattern_slug;
+					$pattern_slugs[ $pattern_slug ] = $pattern_slug_data;
+				}
```
arunshenoy99, line 290:
```suggestion
	public static function get_theme_step_patterns_from_step( $step, $squash = false ) {
		$active_theme = ( \wp_get_theme() )->get( 'TextDomain' );

		if ( ! isset( self::get_theme_step_patterns()[ $active_theme ][ $step ] ) ) {
			return false;
		}

		$pattern_slugs           = self::get_theme_step_patterns()[ $active_theme ][ $step ];
		$block_patterns_registry = \WP_Block_Patterns_Registry::get_instance();

		foreach ( array_keys( $pattern_slugs ) as $pattern_slug ) {
			if ( true !== $pattern_slugs[ $pattern_slug ]['active'] ) {
				continue;
			}

			if ( isset( $pattern_slugs[ $pattern_slug ]['replace'] ) && true === $pattern_slugs[ $pattern_slug ]['replace'] ) {
				$pattern_slug_data = $pattern_slugs[ $pattern_slug ];
				$header_menu_slug = self::get_selected_header_from_flow_data();
				$pattern_slug = ( ! empty( $header_menu_slug ) ) ? $header_menu_slug : $pattern_slug;
				$pattern_slugs[ $pattern_slug ] = $pattern_slug_data;
			}

			$pattern_name = $active_theme . '/' . $pattern_slug;
			if ( ! $block_patterns_registry->is_registered( $pattern_name ) ) {
				continue;
			}

			$pattern = $block_patterns_registry->get_registered( $pattern_name );
				// if header menu slug contains "split" replace the menu links with dummy links
			if ( false !== stripos( $pattern_slug, 'split' ) ) {
				$pattern['content'] = self::replace_split_menu_items( $pattern['content'] );
			}

			if ( ! $squash ) {
				$block_patterns[] = array_merge(
					array(
						'slug'       => $pattern_name,
						'title'      => $pattern['title'],
						'content'    => self::cleanup_wp_grammar( $pattern['content'] ),
						'name'       => $pattern['name'],
						'categories' => $pattern['categories'],
					),
					$pattern_slugs[ $pattern_slug ]
				);
				continue;
			}

			$block_patterns .= self::cleanup_wp_grammar( $pattern['content'] );
		}
```

arunshenoy99 (approved):

PR #221, opened 2023-04-28
PRESS-2 834 Implement Site Classification API (Frontend)

officiallygod:
This branch adds the Frontend part that shows the Primary and Secondary steps from the Site Classification API

https://user-images.githubusercontent.com/48691514/235102148-3e61e08c-9418-476f-8f5c-d2ded2e2b1a1.mp4

PR #237, opened 2023-05-09
Update primary and secondary type data using helpers from the data module

arunshenoy99:
This should be merged only after https://github.com/newfold-labs/wp-module-onboarding/pull/221


abhijitb (approved):
LGTM!

PR #240, opened 2023-05-19
Implement Analytics

arunshenoy99:
Edit: We should be able to pull the 1.0.0 package now.


src/OnboardingSPA/components/App/index.js
```diff
@@ -223,12 +226,57 @@ const App = () => {
...
+		const previousStep = window.nfdOnboarding?.previousStepID;
+		if ( typeof previousStep !== 'string' ) {
+			window.nfdOnboarding.previousStepID = location.pathname;
+			HiiveAnalytics.dispatchEvents();
+			return;
+		}
+
+		if ( previousStep.includes( 'products' ) ) {
+			trackHiiveEvent( 'products-info', {
+				productCount:
```
officiallygod, line 239: Instead of just saving the data that might be in flow, should we be listening if the user clicks and unclicks a certain box? like if many users are likely to select and then deselect a particular option?

src/OnboardingSPA/components/App/index.js
```diff
@@ -223,12 +226,57 @@ const App = () => {
...
+					currentData.storeDetails.productInfo.product_count,
+				productTypes:
+					currentData.storeDetails.productInfo.product_types.join(
+						','
+					),
+			} );
+		}
+
+		if ( previousStep.includes( 'site-pages' ) ) {
+			currentData.data.sitePages?.other?.forEach( ( sitePage ) => {
```
officiallygod, line 249: Same as above step, don't we need to track every click ? and things like does the user even use, description button

src/OnboardingSPA/components/SkipButton/index.js
```diff
@@ -48,6 +49,7 @@ const SkipButton = ( { callback = false } ) => {
 			}
 			setFlow( currentData );
 		}
+		await HiiveAnalytics.dispatchEvents();
```
officiallygod, line 52: Why do we have this here? don't we dispatch on page change?

includes/Services/EventService.php
```diff
@@ -0,0 +1,83 @@
...
+	 * @param array $event The event to sanitize and validate.
+	 * @return array|boolean
+	 */
+	public static function validate( $event ) {
+		if ( ! isset( $event['action'] ) || ! self::validate_action( $event['action'] ) ) {
+			return false;
+		}
+
+		if ( ! isset( $event['category'] ) || ! self::validate_category( $event['category'] ) ) {
+			$event['category'] = Events::get_category();
```
officiallygod, line 78: If a category is not right we are overriding it, so what is the use of the check when we can just use the fixed value itself?

officiallygod (commented):
Some Other Places like:

- Does the user use the Drawer to navigate? 
- How many times the user clicks the help links ( In Page and In the Sidebar )
- Does the user use the Continue Button in the Get Started or Ecom Steps?
- How often does the user go back and forth using the Back and Next page?
- Does the user get the Theme Already Present Modal?
- Does the user encounter an Error Page for Wonder or Woocommerce not installed?
- Does the user somehow magically end up on the Error 404 page?
- Also, are we recording how much time a user spends on a step?

Maybe let's ask David if any of these things should be added or tracked.

src/OnboardingSPA/components/SkipButton/index.js
arunshenoy99: Moved it to inside the `basic-info` check to handle the reload

includes/Services/EventService.php
arunshenoy99: Returning `false` here.

officiallygod (approved):

PR #244, opened 2023-05-29
Refactor onboarding to use the installer module

arunshenoy99:
Lint will fail until we can pull a release `1.0.0` from `newfold-labs/wp-module-installer`


abhijitb (approved):

avneet-raj:
image

image


The Onboarding module is referencing the function from Plugin installer (Inside Bluehost Plugin) which ecommerce module has it in their dependency.

PR #246, opened 2023-05-29
removed hard coded slugs from header menu step

abhijitb:


includes/Data/Patterns.php
```diff
@@ -99,6 +99,7 @@ protected static function get_theme_step_patterns() {
 					'site-header-left-logo-navigation-inline' => array(
 						'active' => true,
 						'shown'  => true,
+						'header' => true,
```
Yashita101, line 102: Instead of adding a new key check, can we have a callback function? Similar to `homepage-styles` in `get_theme_step_filters` ? This way, we would have only the response sent from the backend, and the frontend would display the result based on the header chosen?
Or Can we handle the existence of the key `header` from here itself?
Basically try to send the maximum response from here with minimum alteration to the response in the frontend.

Yashita101 (changes requested):

includes/Data/Patterns.php
abhijitb: This is done. .. thanks

Yashita101 (approved):

PR #259, opened 2023-06-23
PRESS 2 990 Enhance social media validation

officiallygod:


src/OnboardingSPA/components/SocialMediaForm/urlValidator.js
```diff
@@ -0,0 +1,158 @@
+const SocialMediaSites = {
+	FACEBOOK: 'facebook',
```
Yashita101, line 2: can this file be moved to the utils folder?

Yashita101 (changes requested):

src/OnboardingSPA/components/SocialMediaForm/index.js
```diff
@@ -1,235 +1,308 @@
...
+				break;
+		}
+		setSocialData( socialMediaDB );
+	};
+
+	const getErrorType = () => {
+		let error = ERROR_TYPES.NONE;
+		for ( const key in activeErrorTypes ) {
+			if ( activeErrorTypes[ key ] !== ERROR_TYPES.NONE )
+				error = activeErrorTypes[ key ];
```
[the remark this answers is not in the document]
officiallygod: So that was done assuming that the later filled fields that are Tiktok and yelp would have been filled more recently as compared to Facebook and say both have errors in that case the later filled error should be shown rather than the old one. so the error gets replaced with a newer error.

src/OnboardingSPA/components/SocialMediaForm/urlValidator.js
officiallygod: I assume that as this is a very specific helper function for the component SocialMediaForm itself and won't be reused anywhere else it was a more apt location for the file.

abhijitb (approved):

Yashita101 (approved):

src/OnboardingSPA/components/SocialMediaForm/socialMediaModal.js
```diff
@@ -0,0 +1,63 @@
+import { __ } from '@wordpress/i18n';
```
arunshenoy99, line 1: Can we make this a generic `ErrorModal`? `socialMediaModal` is misleading

src/OnboardingSPA/components/SocialMediaForm/socialMediaModal.js
```diff
@@ -0,0 +1,63 @@
...
+		};
+	}, [] );
+
+	const { addNavErrorModalCode, resetNavErrorModal } =
+		useDispatch( nfdOnboardingStore );
+
+	const closeModal = () => {
+		resetNavErrorModal();
+		addNavErrorModalCode( 400 );
+		setShowModal( false );
```
arunshenoy99, line 29: Why are we resetting and then adding a `400` error code and then closing the modal?

src/OnboardingSPA/components/Header/step-navigation.js
```diff
@@ -93,32 +105,42 @@ const Finish = ( { currentData, saveDataAndExitFunc } ) => (
...
 	const isFirstStep = null === previousStep || false === previousStep;
 	const isLastStep = null === nextStep || false === nextStep;
 	return (
 		<div className="nfd-onboarding-header__step-navigation">
 			<ButtonGroup style={ { display: 'flex', columnGap: '0.5rem' } }>
 				{ isFirstStep || isLastStep ? null : (
-					<Back path={ previousStep.path } />
+					<Back
+						path={ previousStep.path }
+						navErrorModalCode={ navErrorModalCode }
```
arunshenoy99, line 131: Pass `overrideNavigation` and check for that in `Back` and `Next`

src/OnboardingSPA/components/SocialMediaForm/urlValidator.js
```diff
@@ -0,0 +1,166 @@
...
+let setErrorsDup;
+let errorTypesDup;
+let setErrorTypesDup;
+
+// Types of errors
+export const ERROR_TYPES = {
+	NONE: 'none',
+	AD_LINK_ERROR: 'ad-link-error',
+	INVALID_LINK_ERROR: 'invalid-link-error',
+};
```
arunshenoy99, line 25: Can we stick to numbers or text either one? I think I saw 400 being set somewhere

arunshenoy99 (changes requested):
@officiallygod some small changes to make

src/OnboardingSPA/components/SocialMediaForm/urlValidator.js
```diff
@@ -0,0 +1,161 @@
...
+	} else {
+		const errorsFiltered = errorsDup.filter( function ( ele ) {
+			return ele !== categ;
+		} );
+		setErrorsDup( errorsFiltered );
+	}
+};
+
+// Common Validation for all links
+const handleCommonValidation = ( categ, url ) => {
```
arunshenoy99, line 37: Let's use combo of `if's and return's` instead of nested `if else's`

arunshenoy99 (approved):

PR #261, opened 2023-06-23
Fix newfold card logo popping up on safari

arunshenoy99:
This one uses an actual HTML element`div` with the background instead of `::before`.


Yashita101 (approved):
LGTM. Thanks!
Also tested the same on Mozilla, Chrome, Microsoft Edge for different devices.

PR #262, opened 2023-06-26
Remove dependency on WooCommerce

arunshenoy99:


includes/Data/Preview.php
```diff
@@ -34,12 +33,9 @@ private static function pre_requisites() {
 				'plugins' => array(),
```
officiallygod, line 33: Maybe we can remove this too as plugin support is not there anymore.

officiallygod (approved):
LGTM!

officiallygod (approved):

PR #264, opened 2023-07-07
Use installer API for theme status

arunshenoy99:
This API doesn't exist in Onboarding now, so this call returns a 404, we have fallbacks in place that throw a dialog to confirm the theme switch, but it's not right to throw the dialog since Wonder should already be active at that point.
Screenshot 2023-07-07 at 2 32 22 PM


officiallygod (approved):
LGTM

PR #268, opened 2023-07-11
Press 2 965 | Refactor Colors step

officiallygod:


src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -43,6 +78,59 @@ const DesignColors = () => {
...
+					block: 'end',
+				} );
+			}
+		}
+	}, [ themeStatus ] );
+
+	/**
+	 * Fetches the color palette for the Drawer and sets the state
+	 * Contains: 'tailored' and 'custom-picker-grouping'
+	 *
```
Yashita101, line 97:
```suggestion
```

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -54,6 +142,14 @@ const DesignColors = () => {
 		}
 	}
 
+	/**
+	 * Converts the user selected value into a suitable valid global styles array value
+	 * param -> selectedColorsLocalTemp -> {base: "#ffffff", contrast: "#404040", ... }
```
Yashita101, line 147: Can the param and return example be stated where the param is declared, line 150,151? Same concern for other functions too.
and have `@return` also documented in the respective functions?

Yashita101 (changes requested):

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: Done

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: Done

Yashita101 (approved):
LGTM

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -43,6 +78,58 @@ const DesignColors = () => {
...
+	 */
+	const getColorStylesAndPatterns = async () => {
+		const globalStyles = await getGlobalStyles();
+		const colorPaletteResponse = await getThemeColors();
+		if ( colorPaletteResponse?.error ) {
+			return updateThemeStatus( THEME_STATUS_INIT );
+		}
+		if ( globalStyles?.error ) {
+			return updateThemeStatus( THEME_STATUS_INIT );
+		}
```
arunshenoy99, line 106: Promise.all?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -43,6 +78,58 @@ const DesignColors = () => {
...
+		setColorPalettes( colorPaletteResponse?.body.tailored );
```
arunshenoy99, line 107: Do we have to check `colorPaletteResponse?.body`? If yes, does it make sense to do it at once instead of 2 times?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -43,6 +78,58 @@ const DesignColors = () => {
...
+		setCustomColorsMap(
+			colorPaletteResponse?.body[ 'custom-picker-grouping' ]
+		);
+		let selectedColorPalette;
+		let selectedColorsLocalTemp;
```
arunshenoy99, line 112: Can we skip the temp variables? I see only one assignment

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -43,6 +78,58 @@ const DesignColors = () => {
...
+	};
+
+	/**
+	 * Helper Function for state in global styles to local mapping
+	 *
+	 * @param {Array} selectedColorPalette - Array of Map structure similar to the one in Global Styles
+	 *                                     e.g. [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+	 *                                     return -> {base: "#ffffff", contrast: "#404040", ... }
+	 */
 	function stateToLocal( selectedColorPalette ) {
```
arunshenoy99, line 133: `stateToLocal` will work only with selected color palettes? If not, does it make sense to move to a utils file? Should a conversion function modify state?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -54,6 +141,14 @@ const DesignColors = () => {
...
 
+	/**
+	 * Converts the user selected value into a suitable valid global styles array value
+	 *
+	 * @param {Object} selectedColorsLocalTemp - Color type mapped to the color
+	 *                                         e.g. {base: "#ffffff", contrast: "#404040", ... }
+	 * @param {string} colorStyle              - Selected Color Palette slug
+	 *                                         e.g. [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+	 */
 	function LocalToState( selectedColorsLocalTemp, colorStyle ) {
```
arunshenoy99, line 152: 1. Camel case? 
2. Should localToState be setting `currentData`? From the description, it seems like it should just convert the value?
3. What is the difference between `local` and `state`?
4. Move to utils and avoid state change here?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -67,16 +162,74 @@ const DesignColors = () => {
...
+	 * @param {Object} colorPalettesTemp - Color Palette from Backend with colors for every palette
+	 * @param {Object} globalStylesTemp  - Global Styles from the store
+	 * @return {Object} selectedGlobalStyle - Updated Global Styles with new color changes
+	 */
 	async function saveThemeColorPalette(
 		colorStyle,
 		colorPalettesTemp = colorPalettes,
-		selectedColorsLocalTemp = selectedColors,
 		globalStylesTemp = storedPreviewSettings
 	) {
```
arunshenoy99, line 234: This one can still be simplified.
1.  Remove all the `temp` variable usage.
2. Change `colorPalettesTemp && colorStyle && selectedThemeColorPalette` to a negation and return.
3. Switch to using `map` instead of modifying the same variable that is being iterated on.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -121,6 +266,100 @@ const DesignColors = () => {
...
+						/>
+						<div
+							className="color-palette__colors--primary"
+							style={ {
+								backgroundColor: `${ colorPalettes[ colorStyle ]?.primary }`,
+							} }
+						/>
+					</div>
+					<div className="color-palette__name drawer-palette--button__text">
+						{ colorStyle?.charAt( 0 ).toUpperCase() +
```
arunshenoy99, line 331: Why use a `?.charAt` when `colorStyle` has already been used for `key`, `onClick` and `onKeyDown`?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -547,7 +656,7 @@ const DesignColors = () => {
 	return (
 		<div className="theme-colors--drawer">
 			<h2>{ __( 'Color Palettes', 'wp-module-onboarding' ) }</h2>
-			{ colorPalettes && buildPalettes() }
+			{ colorPalettes && buildPredefinedPalette() }
```
arunshenoy99, line 659: What's the difference between `predefined` and `default` palette?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -439,7 +548,7 @@ const DesignColors = () => {
 							{ customColors?.base ? <div>&#10003;</div> : null }
 						</div>
 						<div className="custom-palette__below-row-text">
-							Background
+							{ __( 'Background', 'wp-module-onboarding' ) }
```
arunshenoy99, line 551: `SELECT CUSTOM COLORS` and `Reset` are not in `__`

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -526,10 +635,10 @@ const DesignColors = () => {
 				{ showColorPicker && (
 					<Popover>
 						<div
-							className="custom-palette__picker-close-icon"
-							onClick={ () => setShowColorPicker( false ) }
 							role="button"
 							tabIndex={ 0 }
+							className="custom-palette__picker-close-icon"
+							onClick={ () => setShowColorPicker( false ) }
```
arunshenoy99, line 641: What's `#874141` in line 648?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -85,28 +238,20 @@ const DesignColors = () => {
...
-					selectedThemeColorPalette[ idx ].color =
-						selectedColorsLocalTemp[ slug ];
-				} else if (
-					// Add Exception for Background. (perhaps scope to yith-wonder in future)
-					colorPalettesTemp?.[ colorStyle ]?.[ slug ] &&
-					'base' === slug
-				) {
-					selectedThemeColorPalette[ idx ].color = '#FFFFFF';
-				} else if (
 					! isCustomStyle &&
```
arunshenoy99, line 245: Why check for `isCustomStyle`? Shouldn't that happen before determining whether to call `saveThemeColorPalette` vs `saveCustomColors`?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -547,7 +656,7 @@ const DesignColors = () => {
 	return (
 		<div className="theme-colors--drawer">
```
arunshenoy99, line 657: The `colorPalettes` check can move here and be checked only once?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -121,6 +266,100 @@ const DesignColors = () => {
...
+
+	// Custom Colors Section
+
+	/**
+	 * Checks id custom colors has a value
+	 * used to toggle the visiblity of Reset button
+	 *
+	 * @return {boolean} isCustomColorActive
+	 */
+	function isCustomColorActive() {
```
arunshenoy99, line 347: It's better to move this to a state?

arunshenoy99 (commented):
@officiallygod this looks way better than before, the comments help, but let's aim to make the code itself readable. There are still some functions that we can consolidate to reduce a lot of the code

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: Actually did that to save undefined in both ColorPalettes and CustomColors as that will ensure the page doesn't crash else if we don't have colorPaletteResponse.body we would still have to show the page and at least the preview.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: As this only has a use case in the colors file and it will only work with a color palette as almost all the states are different on almost every page.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: So basically state refers to the redux store value that would be saved. So basically the original `state` variable is very limited due to which we created a local version that is basically a smaller and more efficient version of the same, and the function localToState aims to convert it back into the original form after doing faster computations and saving it back to the **state**, hence Local to State.
Again being very specific to the colors step and won't moving it to utils just disperse the files thus reduce readability?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -67,16 +162,74 @@ const DesignColors = () => {
...
+		setSelectedColors( selectedGlobalStylePalette );
+		setSelectedColorsLocal( stateToLocal( selectedGlobalStylePalette ) );
+		trackHiiveEvent( 'color-selection-reset', selectedGlobalStyle.title );
+	}
+
+	/**
+	 * Reset Button to reset selected colors
+	 */
+	async function resetColors() {
+		setToDefaultPalette();
```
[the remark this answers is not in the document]
officiallygod: `setToDefaultPalette` was also being used elsewhere lemme try to combine the two.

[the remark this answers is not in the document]
officiallygod: > Can we stick to only `colors` everywhere instead of a combo of `colorPalette`, `colors` and `colorPalettes`?

Do you mean in JS Docs?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -121,6 +266,100 @@ const DesignColors = () => {
...
+	function buildPredefinedPalette() {
+		return Object.keys( colorPalettes ).map( ( colorStyle, idx ) => {
+			return (
+				<div
+					key={ colorStyle }
+					className={ classNames(
+						'color-palette drawer-palette--button',
+						{
+							'color-palette-selected drawer-palette--button--selected':
+								colorStyle === currentData?.data?.colorStyle,
```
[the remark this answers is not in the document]
officiallygod: This is just to show the highlight and even if `currentData?.data` doesn't exist we have to show the palette and only it won't be highlighted, doing a check above we still have to map and render right?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -1,28 +1,63 @@
...
+	/**
+	 * Custom color mapping selected by the user
+	 * eg: {base: '#da2929', contrast: '#404040',.. }
+	 */
 	const [ customColors, setCustomColors ] = useState();
-	const [ colorPalettes, setColorPalettes ] = useState();
+	/**
+	 * Determines which custom color was toggled to select a color
+	 * e.g. base, primary, secondary..
+	 */
```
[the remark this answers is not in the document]
officiallygod: Sure

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: Nothing, Just changed that to distinguish between Custom and the one we we have already defined, Can change it to default.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: That is the default value for the color picker when opened, will remove that.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -180,73 +427,20 @@ const DesignColors = () => {
...
-		trackHiiveEvent( 'color-selection', colorStyle );
-	};
-
+	/**
+	 * Change the color for the active color picker
+	 * this changes the color when the user clicks on a new color.
+	 *
+	 * @param {string} color - The Color hex user has selected
+	 */
 	const changeCustomPickerColor = async ( color ) => {
```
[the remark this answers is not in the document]
officiallygod: So **selectCustomColor** is run when we click on the color like Background and to open and close the color picker this is used.
**changeCustomPickerColor** runs independently as it knows who toggled the color picker and saves that to the store.

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -121,6 +266,100 @@ const DesignColors = () => {
...
+	}
+
+	/**
+	 * Find the index of the color type being changed in the array
+	 * Array to check -> [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+	 *
+	 * @param {string} slugName - Slug Name for color i.e. base, secondary, header-title,...
+	 * @return {number} res - Index of the key "slug" mapped ({color: '#ffffff', name: 'Base', slug: 'base'}, ...) in the array
+	 */
 	function findInCustomColors( slugName ) {
```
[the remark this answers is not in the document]
officiallygod: So basically if a wrong mapping is done in backend say `base -> 'headwer-title'` then it will search for index of **'headwer-title'** in palette but when that won't be there it will return the index of '**base**'

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
officiallygod: Considering its use case won't the state be more memory-consuming rather this? as the state would update causing a re-render every time we move the color picker by one color?

src/OnboardingSPA/components/Drawer/DrawerPanel/DesignColors.js
```diff
@@ -121,6 +266,100 @@ const DesignColors = () => {
...
+	 * Select a color from predefined palette and sync it.
+	 *
+	 * @param {string} colorStyle - Selected Color Palette slug
+	 */
+	const handleClick = ( colorStyle ) => {
+		if ( selectedColors?.slug === colorStyle ) {
+			return true;
+		}
+
+		clearCustomColors();
```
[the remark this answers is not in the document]
officiallygod: So basically this ensures that the custom colors reflect the same colors in the predefined palette that was clicked.

src/OnboardingSPA/components/ColorPickerButton/index.js
```diff
@@ -0,0 +1,33 @@
+// This is an Color Picker Option with a color displayed infront of the name
```
arunshenoy99, line 1: I think we can remove this comment/follow the JS standard.

src/OnboardingSPA/components/ColorPickerButton/index.js
```diff
@@ -0,0 +1,33 @@
...
+const ColorPickerButton = ( {
+	isColorSelected,
+	color,
+	slug,
+	name,
+	callback,
+} ) => {
+	return (
+		<div
+			className="custom-palette__below-row"
```
arunshenoy99, line 11:
```suggestion
			className="color-picker-button__row"
```
Similarly for the other ones.

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * Helper Function for state in global styles to local mapping
+ *
+ * @param {Array} selectedColorPalette - Array of Map structure similar to the one in Global Styles
+ *                                     e.g. [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+ *                                     return -> {base: "#ffffff", contrast: "#404040", ... }
+ */
+export function storeToState( selectedColorPalette ) {
```
arunshenoy99, line 8:
```suggestion
export function storeToState( selectedColorsInStore ) {
```

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+	if ( selectedColorPalette ) {
+		const selectedColorsLocalTemp = {};
```
arunshenoy99, line 10:
```suggestion
		const selectedColorsInState = {};
```

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Converts the user selected value into a suitable valid global styles array value
+ *
+ * @param {Object} selectedColorsLocalTemp - Color type mapped to the color
+ *                                         e.g. {base: "#ffffff", contrast: "#404040", ... }
+ * @param {string} colorStyle              - Selected Color slug
+ *                                         e.g. [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+ */
+export function stateToStore( selectedColorsLocalTemp, colorStyle ) {
```
arunshenoy99, line 26:
```suggestion
export function stateToStore( selectedColorsInState, colorStyle ) {
```
Also do we actually need `colorStyle`, I can see it is part of the `if` but it is not being used anywhere.

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+	if ( selectedColorsLocalTemp && colorStyle ) {
+		const colorsArray = [];
```
arunshenoy99, line 28:
```suggestion
		const selectedColorsInStore = [];
```

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @param {string} slugName              - Slug Name for color i.e. base, secondary, header-title,...
+ * @param {Object} storedPreviewSettings - The Stored Preview Settings for live preview
+ * @param {string} colorPickerCalledBy   - The Color that triggered the color picker
+ * @return {number} res - Index of the key "slug" mapped [{color: '#ffffff', name: 'Base', slug: 'base'}, ...] in the array
+ */
+export function findInPalette(
+	slugName,
+	storedPreviewSettings,
+	colorPickerCalledBy
```
arunshenoy99, line 54:
```suggestion
	slugName,
	fallbackSlugName
	palette,
```
And we can pass the `palette` to the function.

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+) {
+	const selectedThemeColorPalette =
+		storedPreviewSettings?.settings?.color?.palette;
+	const res = selectedThemeColorPalette.findIndex(
```
arunshenoy99, line 58:
```suggestion
	const index = selectedThemeColorPalette.findIndex(
```

src/OnboardingSPA/pages/Steps/DesignColors/utils.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Finds the index of the color type being changed in the array
+ * Array to check -> [{color: '#ffffff', name: 'Base', slug: 'base'}, ...]
+ *
+ * @param {string} slugName              - Slug Name for color i.e. base, secondary, header-title,...
+ * @param {Object} storedPreviewSettings - The Stored Preview Settings for live preview
+ * @param {string} colorPickerCalledBy   - The Color that triggered the color picker
+ * @return {number} res - Index of the key "slug" mapped [{color: '#ffffff', name: 'Base', slug: 'base'}, ...] in the array
+ */
+export function findInPalette(
```
arunshenoy99, line 51:
```suggestion
export function findIndexInPalette(
```

arunshenoy99 (approved):
LGTM, thanks @officiallygod

PR #274, opened 2023-08-03
Add/Update analytics events. To check if the approach is correct.

diwanshuster:
WIP on analytics events refactor.


arunshenoy99 (commented):
Unfortunately, the approach of modifying existing `trackHiiveEvent` calls will not work, nor meet the requirements discussed.

PR #288, opened 2023-08-10
Expedite theme install if wp-cron fails to install in time

arunshenoy99:
It seems like WP_CRON is too slow on production sites and are causing the task to be stuck in the queue for a long time. This is causing the design steps to show the confirmation dialog. This PR expedites the installation task once a user reaches the design steps and the theme has not been installed yet.


officiallygod (approved):
LGTM!

PR #293, opened 2023-08-23
Enhance Translations and Generate POT file

arunshenoy99:


officiallygod (approved):
LGTM!

wpscholar:
@arunshenoy99 - Have you seen the annotations on this page? https://github.com/newfold-labs/wp-module-onboarding/actions/runs/5996833546

It shows a number of things we may need to address with regard to the React `useEffect` dependencies

arunshenoy99:
> @arunshenoy99 - Have you seen the annotations on this page? https://github.com/newfold-labs/wp-module-onboarding/actions/runs/5996833546
> 
> It shows a number of things we may need to address with regard to the React `useEffect` dependencies

I have @wpscholar, those warnings all point to functions that will be available to the component. We have a task for looking into warnings, but not a priority. For this particular case, we would have to memoize those functions before passing them as a dependency to `useEffect`, but as of now it's not really worth doing other than for fixing the lint warning, like I mentioned, the function would always be available.

I also noticed similar usages across the Bluehost Plugin as well that would throw this warning. Linking 2 of them, but can see many such instances.  Are we planning on running a JS lint for the BH Plugin to catch these?
https://github.com/bluehost/bluehost-wordpress-plugin/blob/main/src/app/data/store.js
https://github.com/bluehost/bluehost-wordpress-plugin/blob/main/src/app/pages/marketplace/index.js

PR #301, opened 2023-09-06
Grid Component

officiallygod:


src/OnboardingSPA/components/GridView/index.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Common Grid View Component
+ * Shows the components in a grid view
+ * Supports column size 2 - 5
+ *
+ * @return GridView of children
+ */
+
+const GridView = ( { size, colGap = 0, children } ) => {
```
arunshenoy99, line 9: Let's add a custom `className` prop as well? We can concatenate with `grid-basic grid-col-${ size }` using `classnames`. I'm thinking this will give consumers better control over the component. Also, should we name it `Grid` instead of `GridView`?

src/OnboardingSPA/components/GridView/stylesheet.scss
```diff
@@ -0,0 +1,29 @@
+.grid-basic {
```
arunshenoy99, line 1: Should we add an onboarding prefix to the CSS selectors? This will ensure that it applies only this component, given that `grid` is a very common name.

officiallygod: Sure, Thanks

src/OnboardingSPA/components/GridView/index.js
officiallygod: Sure, Thanks

arunshenoy99 (approved):

PR #304, opened 2023-09-07
Redirect all users to the brand plugin dashboard

arunshenoy99:


src/OnboardingSPA/components/SkipButton/index.js
```diff
@@ -7,7 +7,7 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import { setFlow } from '../../utils/api/flow';
 import { store as nfdOnboardingStore } from '../../store';
 import { getSettings, setSettings } from '../../utils/api/settings';
-import { wpAdminPage, pluginDashboardPage } from '../../../constants';
```
officiallygod, line 10: Can we get rid of `wpAdminPage` from constants in total as I don't see it being used anywhere and we are just doing `addQueryArgs` on it for no reason?

officiallygod (commented):
Rest of the code looks good to me.

src/OnboardingSPA/components/SkipButton/index.js
arunshenoy99: I still think we should leave it as a fallback.

officiallygod (approved):

PR #328, opened 2023-10-10
Add defensive checks for edge cases

arunshenoy99:
1. When the GET settings API cannot read the corrupted `wpseo_social` option and fails, it tends to return a failed API response. We then attempt to read this false value in React, causing a white screen.
2. Point 1 leads to a false value being reported to the POST API, triggering an error in the debug log. While this doesn't result in a white screen, as a best practice, we have added a check anyway.

Both of these issues are incredibly rare but were prioritized for resolution. They are also challenging to replicate because of the rare nature.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -483,6 +483,76 @@
             "time": "2023-01-05T11:28:13+00:00"
         },
         {
+            "name": "doctrine/instantiator",
+            "version": "1.5.0",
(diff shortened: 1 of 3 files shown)


diwanshuster (approved):
LGTM

officiallygod (approved):

PR #329, opened 2023-10-10
Cypress Tests events API for get-experience page

avneet-raj:
Added Cypress Tests for events API on get-started-experience step for different API calls being made at the time of different options selected based on experience level.

### DIFF
--- a/tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
+++ b/tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
@@ -7,6 +7,7 @@ import {
 	CheckInfoPanel,
 	CheckIntroPanel,
 } from '../wp-module-support/sidebar.cy';
+import { APIList } from '../wp-module-support/EventsApi.cy';
 
 describe( 'Start Setup WP Experience Page', function () {
 	before( () => {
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
```diff
@@ -65,6 +65,45 @@ describe( 'Start Setup WP Experience Page', function () {
 		} );
 	} );
 
+	const EventsAPI = ( experience_val ) => {
+		cy.intercept('http://localhost:8882/index.php?rest_route=%2Fnewfold-onboarding%2Fv1%2Fevents%2Fbatch&flow=wp-setup&_locale=user').as('events');
+		cy.wait('@events').then((interception) => {
+			const responseBody = interception.request.body;
+			const responseData1 = responseBody[0].data;
+			const responseData2 = responseBody[1].data;
```
arunshenoy99, line 73: What is the difference between `responseData1` and `responseData2`? Do we require both of them?

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
```diff
@@ -65,6 +65,45 @@ describe( 'Start Setup WP Experience Page', function () {
...
+			};
+		});
+	};
+
+	it( 'Check if events API call being made after radio buttons are clicked', () => {
+		let radioCount = 0;
+		const className = '.components-radio-control__option';
+		const arr = cy.get( className );
+		arr.each( () => {
+			cy.reload();
```
arunshenoy99, line 88: Do we need to `reload` the page every time we click a radio button? That's not in line with how the user will be clicking the buttons.

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
```diff
@@ -65,6 +65,45 @@ describe( 'Start Setup WP Experience Page', function () {
...
+			cy.wait(2000);
+			cy.get( '[type="radio"]' )
+				.eq( radioCount )
+				.click( { force: true } )
+				if(radioCount==0){
+					EventsAPI("novice");
```
arunshenoy99, line 94: Does it make sense to store this in a `data.json` file and map `0, 1, 2` to `novice, intermediate, expert`?

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
avneet-raj: This is needed for the edge case, sometimes the experience _level comes into the second data and sometimes in the first. I've refined it more now so that, it doesn't fail. 90% of the time it'll come in the resposeData1.

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
avneet-raj: Yes, this'll be needed to clear the older network logs as every time when a radio button is clicked, the API call is overlapped on the earlier one (happening only on cypress). So, to eliminate this, we have to use reload before every radio button click.

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
avneet-raj: I think, we can keep this over here only. We'll have other variables for other steps as well, so I guess it'll make more confusing to use variables like these in a single file.

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
```diff
@@ -65,6 +65,45 @@ describe( 'Start Setup WP Experience Page', function () {
 		} );
 	} );
 
+	const EventsAPI = ( experience_val ) => {
+		cy.intercept('http://localhost:8882/index.php?rest_route=%2Fnewfold-onboarding%2Fv1%2Fevents%2Fbatch&flow=wp-setup&_locale=user').as('events');
+		cy.wait('@events').then((interception) => {
```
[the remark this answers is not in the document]
avneet-raj: yes

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
```diff
@@ -65,6 +65,45 @@ describe( 'Start Setup WP Experience Page', function () {
 		} );
 	} );
 
+	const EventsAPI = ( experience_val ) => {
+		cy.intercept('http://localhost:8882/index.php?rest_route=%2Fnewfold-onboarding%2Fv1%2Fevents%2Fbatch&flow=wp-setup&_locale=user').as('events');
```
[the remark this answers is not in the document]
avneet-raj: Yes, for now I have moved all the API in a single file. I'll do the regex that matches the URL also.

tests/cypress/integration/2-general-onboarding-flow/get-started-experience.cy.js
avneet-raj: Removed reload.

PR #340, opened 2023-10-18
Update path to check lint of cypress tests as well

avneet-raj:
Updated the path of GitHub workflow Lint-Check-SPA to check Lint of cypress tests as well.


avneet-raj:
image

PR #347, opened 2023-10-31
Cypress api events for site type secondary page

avneet-raj:


ajayadav09 (commented):
checked the assertions on local @avneet-raj  will you also be adding the regex pattern for baseUrl in this PR

arunshenoy99 (approved):

ajayadav09 (approved):

PR #362, opened 2023-11-10
Press2 1233 merge components aionboarding

ajayadav09:


src/OnboardingSPA/components/ThemeContextProvider/index.js
```diff
@@ -0,0 +1,21 @@
+import { useState, createContext } from '@wordpress/element';
+
+const ThemeContext = createContext();
+
+const ThemeProvider = ( { children } ) => {
+	const [ theme, setTheme ] = useState( 'dark' );
+
+	const toggleTheme = () => {
+		setTheme( ( prevTheme ) =>
+			prevTheme === 'dark' ? 'light' : 'dark'
```
officiallygod, line 10: Should we have constants? for `dark` and `light` cause I see the use case a lot.
I remember we had `Background` earlier and one day that was changed to `Base`

src/OnboardingSPA/components/AdminBar/index.js
```diff
@@ -1,7 +1,40 @@
...
 		<div className="nfd-onboarding-header__admin-bar">
-			Admin Bar Goes Here
+			<div className="nfd-onboarding-header__admin-bar__wplogo">
+				<Icon icon={ wordpress } />
+				<span>{ __( 'WordPress', 'wp-module-onboarding' ) }</span>
+			</div>
+			<div className="nfd-onboarding-header__admin-bar__profile">
+				<span className="nfd-onboarding-header__admin-bar__profile__greeting">
+					<span>
+						{ __( 'Howdy!', 'wp-module-onboarding' ) },{ ' ' }
```
officiallygod, line 27: Do we have an exclamation mark? Also, let's not use empty strings like that. Can't it be part of the same string?

officiallygod (commented):
Can you please check the `Normal flow` once? I think it is breaking.

src/OnboardingSPA/components/AdminBar/index.js
ajayadav09: yes there is an exclamation, will remove the space

ajayadav09:
Fixed the normal flow class name issue by adding the current step and current flow check

officiallygod (changes requested):
I think the Light Mode button is different

Screenshot 2023-11-17 at 5 06 11 PM

officiallygod (approved):

PR #374, opened 2023-11-28
Update logic for onboarding redirects

wpscholar:
This PR utilizes the new Install Checker module to check if a site is a fresh WordPress install and updates the logic for redirects as follows: 

- Always skip onboarding if user isn't an admin
- Allow manually disabling onboarding (via query param - sets DB option)
- Allow manually enabling onboarding (via query param - sets DB option)
- Check if onboarding is explicitly on or off (via DB option)
- If explicitly on, go ahead and trigger onboarding
- If explicitly off, bypass onboarding
- If not explicitly set, run through qualifiers
  - User hasn't completed or exited onboarding already
  - Site is fresh install (use module for this)
  - Coming soon is not enabled
- If onboarding redirect is triggered at any point, update the DB option to disable onboarding for future visits

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
 		"mustache/mustache": "^2.14",
 		"wp-cli/wp-config-transformer": "^1.3",
 		"newfold-labs/wp-module-onboarding-data": "^0.0.7",
-		"newfold-labs/wp-module-patterns": "^0.1.8"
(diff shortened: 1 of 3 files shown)


arunshenoy99:
Thanks @wpscholar, have included these as part of https://github.com/newfold-labs/wp-module-onboarding/pull/373. Hence, closing out this PR.

PR #376, opened 2023-11-29
Minor dark mode fixes and slot filled footers

arunshenoy99:
This PR makes minor fixes to the positioning and animation of the dark mode button. Also, it introduces slot-filled footers, this should greatly simplify the implementation of various step-specific footer layouts in the new v2 Figma.


officiallygod (approved):
LGTM

PR #391, opened 2023-12-11
Mark AI Sitegen Generated Themes on themes.php

arunshenoy99:
image

### DIFF
--- a/src/Scripts/sitegen-theme-marker/sitegen-theme-marker.js
+++ b/src/Scripts/sitegen-theme-marker/sitegen-theme-marker.js
@@ -1,0 +1,21 @@
+window.onload = function () {
+	const homepages = window.nfdOnboarding.homepages.data;
+	const activeTheme = window.nfdOnboarding.active;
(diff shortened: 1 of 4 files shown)


src/Scripts/sitegen-theme-marker/sitegen-theme-marker.js
```diff
@@ -0,0 +1,21 @@
+window.onload = function () {
+	const homepages = window.nfdOnboarding.homepages.data;
```
officiallygod, line 2: Should we add a `?` in case the user comes to the themes page without completing Onboarding?

officiallygod (approved):
LGTM!

src/Scripts/sitegen-theme-marker/sitegen-theme-marker.js
arunshenoy99: This will always be in an empty array from the beginning. In the backend, we have a backup check implemented as well `( ! $flow_data || ! isset( $flow_data['sitegen']['homepages'] ) )`

PR #392, opened 2023-12-12
Fix onboarding flow

wpscholar:
Update onboarding to only disable after a successful API fetch when onboarding has loaded.


circlecube (approved):
Looks good

PR #401, opened 2024-01-04
Create Brand Plugin Cypress Tests GitHub Workflow in Module

avneet-raj:
Create a workflow to run Tests for Module in Brand Plugin


circlecube:
Now that the postcss config file is no longer needed here the package updates are not required. I think they're a good idea to keep things updated, but if that is a hold up to merging this workflow in, we can lose the package updates completely. I'm happy to revert those if that would help.

arunshenoy99:
> Now that the postcss config file is no longer needed here the package updates are not required. I think they're a good idea to keep things updated, but if that is a hold up to merging this workflow in, we can lose the package updates completely. I'm happy to revert those if that would help.

Thanks @circlecube. I agree that it's essential that we keep them updated. We'll test Onboarding with the new versions, and if nothing breaks, we'll merge this one along with the package updates, otherwise, we'll revert them and merge only the workflow.

officiallygod (approved):
LGTM!

PR #407, opened 2024-01-11
Fix fonts for orchid shades and correct typography step issues

arunshenoy99:
1. Fixes issue with fonts not applying to the orchid shades preview.
2. Fixes the typography step that incorrectly displays the Finish button on a page refresh.


officiallygod (approved):
Looks good to me.

PR #411, opened 2024-01-12
PRESS2-1485 Cypress Tests for SiteGen Welcome Step

avneet-raj:


tests/cypress/integration/5-AI-SiteGen-onboarding-flow/welcome.cy.js
```diff
@@ -0,0 +1,59 @@
...
+    } );
+
+    it( 'Check the Get Started button', () => {
+        cy.get( '.nfd-onboarding-button--site-gen-next' )
+            .should('be.visible')
+            .should('have.text','Get Started')
+            .click();
+        cy.wait(2000);
+        cy.url().should('not.contain', 'sitegen/step/welcome');
+    } );
```
sangeetha-nayak, line 58: Add an assertion to verify if next step is site-detail step or not.

sangeetha-nayak (changes requested):

tests/cypress/integration/5-AI-SiteGen-onboarding-flow/welcome.cy.js
avneet-raj: No, we don't require this. As, in future, if we have the order changes, the test will start to fail. So, this will be sufficient to test that we jump onto the next step.
We have used the same approach in the earlier onboarding flow also.

sangeetha-nayak (approved):

PR #425, opened 2024-01-19
Sitegen: Code Refactor and Post-Integration Fixes

arunshenoy99:
https://github.com/newfold-labs/wp-module-onboarding-data/pull/33


includes/RestApi/SiteGenController.php
```diff
@@ -210,45 +191,43 @@ public function generate_sitegen_meta( \WP_REST_Request $request ) {
...
 		}
-		if ( ! $content_style || is_wp_error( $content_style ) ) {
+
+		$content_style = SiteGenService::instantiate_site_meta( $site_info, 'content_tones' );
+		if ( is_wp_error( $content_style ) ) {
 			return new \WP_Error(
 				'nfd_onboarding_error',
 				__( 'Required data is missing.', 'wp-module-onboarding' ),
 				array( 'status' => 400 )
 			);
```
officiallygod, line 217: The return value is itself a `wp_error` should we just return that? if either `$target_audience || $content_style` is an `is_wp_error`? 

Also I was wondering if we mention in instantiate_site_meta which key failed as we can never know in this case until we check the db.

includes/RestApi/SiteGenController.php
```diff
@@ -210,45 +191,43 @@ public function generate_sitegen_meta( \WP_REST_Request $request ) {
 	 * @return array
 	 */
 	public function get_homepages( \WP_REST_Request $request ) {
+		$existing_homepages = SiteGenService::get_homepages();
+		if ( ! empty( $existing_homepages ) ) {
+			return new \WP_REST_Response( $existing_homepages, 200 );
+		}
```
officiallygod, line 197: Will this work? Cause I believe we do check in flow for `homepages > data` in flow before we make this call and won't this result in an unnecessary wp read?

package.json
```diff
@@ -20,7 +20,7 @@
         "webpack-merge": "^5.8.0"
     },
     "scripts": {
-        "build": "wp-scripts build ./src/onboarding.js",
+        "build": "wp-scripts build ./src/onboarding.js ./src/Scripts/sitegen-theme-marker/sitegen-theme-marker.js ./src/Scripts/sitegen-theme-marker/sitegen-theme-marker.css",
```
officiallygod, line 23: Just wondering if this is the right place or is it adding an entry point in `webpack.config.js` ?

officiallygod (commented):
Screenshot 2024-01-25 at 10 24 05 AM
Screenshot 2024-01-25 at 10 24 13 AM
Screenshot 2024-01-25 at 10 25 00 AM

https://github.com/newfold-labs/wp-module-onboarding/assets/48691514/a771592e-241b-4800-b18d-6f691a9ab79a

Might need a bit more sync?

officiallygod (approved):
Looks good

PR #436, opened 2024-01-30
UI bug fixes reported

ajayadav09:
Refactored the center component and seperated the title component and the drop down component. It needed a local state to type in wothout re-rendering and added useEffect to update on the basis of that local state and change homepageTitle when content re-generates.


src/OnboardingSPA/steps/SiteGen/Editor/Header/center.js
```diff
@@ -140,14 +123,89 @@ const StepEditorHeaderCenter = ( {
...
+		}
+	};
+
+	const onRenameItemSelect = () => {
+		enableInput( true );
+	};
+
+	const onRename = () => {
+		if ( typeof handleRename === 'function' ) {
+			handleRename( inputRef.current.value );
```
officiallygod, line 189: Should we change the data in homepages if rename was selected and never edited? That does cause an unnecessary re-render. Also should we trim the data?

src/OnboardingSPA/steps/SiteGen/Preview/stylesheet.scss
```diff
@@ -55,6 +55,8 @@
 		margin: 10px 10px 55px 10px;
 		flex-wrap: wrap;
 		max-width: 1440px;
+		align-items: center;
+		justify-content: center;
```
officiallygod, line 59: Screenshot 2024-01-31 at 7 15 14 PM
Screenshot 2024-01-31 at 7 15 47 PM

Doesn't look right 👀

ajayadav09: @officiallygod  Have updated the code with by adding width to the parent container. The width will be different for devices.

officiallygod (approved):
Looks good to me!

officiallygod (approved):

PR #438, opened 2024-01-30
Separate out header, content and footer for homepages

arunshenoy99:
https://github.com/newfold-labs/wp-module-onboarding-data/pull/35

### DIFF
--- a/src/OnboardingSPA/steps/SiteGen/Editor/index.js
+++ b/src/OnboardingSPA/steps/SiteGen/Editor/index.js
@@ -81,9 +81,16 @@ const StepSiteGenEditor = () => {
 				populateFontsInPreviewSettings( newPreviewSettings );
 		}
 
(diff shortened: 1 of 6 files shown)


src/OnboardingSPA/steps/SiteGen/Editor/index.js
```diff
@@ -81,9 +81,16 @@ const StepSiteGenEditor = () => {
 				populateFontsInPreviewSettings( newPreviewSettings );
 		}
 
+		let blockGrammar = '';
+		[ 'header', 'content', 'footer' ].forEach( ( part ) => {
```
officiallygod, line 85: I see this array is used in a lot of places, should we move this to a constant or a util func that combines and gives the Block Grammar?

officiallygod (approved):
Looks good to me other than that.

PR #439, opened 2024-01-30
relaced the gif with orb animation

ajayadav09:
Used the compiled CSS of the orb logo from the https://github.com/newfold-labs/css-wonder-animation

https://jira.newfold.com/browse/PRESS2-1630


officiallygod (commented):
Should we be adding the [package](https://github.com/newfold-labs/css-wonder-animation) itself? As won't it be difficult to keep track of changes in their end without releases and integrate them here?

PR #462, opened 2024-02-12
Save continue dim UI

ajayadav09:
add set pointer events to none for the body on Save and Continue


src/OnboardingSPA/steps/SiteGen/Editor/Header/stylesheet.scss
```diff
@@ -160,3 +160,15 @@
 		}
 	}
 }
+
+body {
+
+	&.editor-actions-disabled {
+		pointer-events: none;
+		opacity: 0.7;
```
officiallygod, line 168: Will this even fade out the Save and Continue button itself?

ajayadav09: yes this fades out the Save and Continue button button too but it is not visually opaque and achieves the effect or alternatively we can use an overlay.

src/OnboardingSPA/steps/SiteGen/Editor/Header/index.js
```diff
@@ -222,6 +222,26 @@ const StepSiteGenEditorHeader = () => {
...
+			}
+		}
+
+		return () => {
+			if ( editorElement ) {
+				editorElement.classList.remove( 'editor-actions-disabled' );
+			}
+		};
+	}, [ isSaving ] );
+
```
arunshenoy99, line 244: We should create a state, something like `interactionDisabled`, and disable them at the parent level component using a `useEffect` instead of selecting the element by a class name, which can change with any WordPress release. We can make this a generic overlay state that can be used across any step in onboarding to prevent interaction. Let's also think about what we need to do incase we need to switch to an error state when something fails.

src/OnboardingSPA/steps/SiteGen/Editor/Header/index.js
```diff
@@ -319,6 +339,7 @@ const StepSiteGenEditorHeader = () => {
 					</div>
 				</div>
 			</Fill>
+			<div className="nfd-editor-actions-overlay"></div>
```
arunshenoy99, line 342: Use `nfd-onboarding-*` as a prefix.

src/OnboardingSPA/steps/SiteGen/Editor/Header/stylesheet.scss
```diff
@@ -160,3 +160,26 @@
 		}
 	}
 }
+
+.nfd-interface-interface-skeleton__editor {
```
arunshenoy99, line 164: Leverage SASS capabilities and move `__editor` inside.

src/OnboardingSPA/steps/SiteGen/Editor/Header/stylesheet.scss
```diff
@@ -160,3 +160,26 @@
...
+			position: fixed;
+			top: 0;
+			left: 0;
+			right: 0;
+			bottom: 0;
+			background: rgba(var(--nfd-onboarding-primary-rgb), 0.3);
+			z-index: 1;
+		}
+
+		.nfd-onboarding-header--sitegen__editor__end__save-button {
```
arunshenoy99, line 178: Leverage SASS capabilities to break down `--`'s and `__`'s.

src/OnboardingSPA/steps/SiteGen/Editor/Header/stylesheet.scss
```diff
@@ -160,3 +160,26 @@
...
+			position: relative;
+			z-index: 2;
```
arunshenoy99, line 180: Is there a specific reason why we have a `z-index: 2;` here? Does this button need to be on top for any particular purpose?

src/OnboardingSPA/steps/SiteGen/Editor/Header/stylesheet.scss
```diff
@@ -160,3 +160,26 @@
 		}
 	}
 }
+
+.nfd-interface-interface-skeleton__editor {
+
+	&.editor-actions-disabled {
+
+		.nfd-editor-actions-overlay {
```
arunshenoy99, line 168: Can we also add `cursor: not-allowed;` so that it's clear that no interaction is allowed?

arunshenoy99 (commented):
Can we fix the SPA lint workflow as well?

ajayadav09:
Screenshot 2024-02-15 at 5 05 36 PM

Screenshot 2024-02-15 at 5 06 19 PM

Light mode and dark mode

PR #463, opened 2024-02-13
Bug Fixes

officiallygod:
1. Fixes The Fork step showing up in Drawer.
2. Facebook step mobile view 'next' issue.
3. Fixes regenerate overflowing in Preview Step.
4. Adds the text change and removal of skip in Experience Step.

Screenshot 2024-02-19 at 6 34 33 PM


src/OnboardingSPA/components/Drawer/DrawerPanel/NavPrimary.js
```diff
@@ -27,6 +28,9 @@ const NavPrimary = () => {
 			<div className="nfd-onboarding-drawer__panel-menu">
 				<ul className="nfd-onboarding-drawer__panel-routes">
 					{ topSteps.map( ( step ) => {
+						if ( step === stepTheFork ) {
+							return false;
+						}
```
arunshenoy99, line 33: We should define a key in the `step.js` for the fork step, such as `drawerNavigation: false;`, and then implement the necessary checks wherever needed.

src/OnboardingSPA/components/CardWithOptions/index.js
```diff
@@ -1,7 +1,8 @@
+import classNames from 'classnames';
 import { memo } from '@wordpress/element';
 import { Icon, chevronRight } from '@wordpress/icons';
 
-const CardWithOptions = ( { title, options, skip, callback } ) => {
+const CardWithOptions = ( { title, options, skip, selection, callback } ) => {
```
arunshenoy99, line 5: We could consider renaming `CardWithOptions` to something like `OptionWithTitleAndDescription` and pass `isSelected` directly to each specific option instead of having a generic `CardWithOptions` component that iterates over `options` in a particular format. This approach would provide us with better control over the component and enhance its reusability.

So in this case, the responsibility of looping over options and rendering them could be moved to the experience step or another appropriate card component, separating them and making each component more focused on its specific task.

src/OnboardingSPA/steps/SiteGen/Experience/index.js
```diff
@@ -48,25 +44,27 @@ const SiteGenExperience = () => {
...
-		}
 	};
 
 	return (
 		<SitegenAiStateHandler>
 			<CommonLayout isCentered>
 				<div className={ 'nfd-sg-experience-level' }>
-					<SiteGenLoader />
+					<SiteGenLoader
+						watcher={ selection !== 0 ? true : false }
```
arunshenoy99, line 59: I don't think we should autonavigate if a selection was not manually made by the user, i.e., via a click event.

src/OnboardingSPA/components/Loaders/SiteGenLoader/index.js
```diff
@@ -4,7 +4,7 @@ import { useNavigate } from 'react-router-dom';
 import { memo, useEffect, useState } from '@wordpress/element';
 import { store as nfdOnboardingStore } from '../../../store';
 
-const SiteGenLoader = ( { autoNavigate = false } ) => {
+const SiteGenLoader = ( { customNavPercentage, watcher = null, autoNavigate = false } ) => {
```
arunshenoy99, line 7: Could we avoid using `customNavPercentage` and let it increment naturally with the steps?

arunshenoy99 (commented):
https://github.com/newfold-labs/wp-module-onboarding/assets/38878906/50fbc08e-7a0b-40f9-bb76-0a726ff43b90

src/OnboardingSPA/components/Loaders/SiteGenLoader/index.js
officiallygod: This is to ensure that only when it is 50 navigate to the next step (building -experience) and when 100 navigate to next (building - Preview)

src/OnboardingSPA/steps/SiteGen/Experience/index.js
officiallygod: Let me remove that then.

src/OnboardingSPA/components/Drawer/DrawerPanel/NavPrimary.js
officiallygod: Sure

officiallygod:
> auto-proceeding.mov

I think this is because its's cached, so if we ever make a choice then we won't be asked again.

src/OnboardingSPA/components/CardWithOptions/index.js
officiallygod: Ohhh Can you tell me a bit more about this, Didn't understand completely as I believe those individual options won't be reusable as they will have a specific styling to them tightly coupled and we would need the outer box styling wherever we reuse it.

arunshenoy99 (approved):
LGTM thanks!

PR #484, opened 2024-02-22
Update to use Node 16 instead of 14

wpscholar:
Update to use Node 16 instead of 14 (need to move to 20).

- Adds an `.nvmrc` file to define Node version
- Update GitHub workflow to use Node 16
- Update `package-lock.json` file for Node 16

At a minimum, moving to Node 16 means we don't have to jump through hoops to run the build locally on M1 or M2 chips.


arunshenoy99 (approved):
Thanks, @wpscholar, moving this for a round of QA before merging.

PR #490, opened 2024-02-27
updated fork: redirect to old flow if ai sitegen is false

girish-lokapure:


src/OnboardingSPA/components/StartOptions/index.js
```diff
@@ -30,6 +30,7 @@ const StartOptions = ( { questionnaire, oldFlow, options } ) => {
 		if ( ! validateFlow( brandConfig, newFlow ) ) {
 			return false;
 		}
+		window.nfdOnboarding.currentBrand = brandConfig;
```
arunshenoy99, line 33: I don't think it's safe to store brandConfig in the window object.

officiallygod (approved):

PR #500, opened 2024-02-29
Skip Fork Screen if Sitegen Capability is disabled and fix back button issue

girish-lokapure:
-  added capability check in state handler, removes the fork step if the flow is ecom or sitebuild
-  this fixes the back button issue as well


ajayadav09 (commented):
Tested. works as expected

diwanshuster (approved):

PR #515, opened 2024-03-07
Fix sitegen editor fonts loading issue

mr-vara:
This pull request (PR) aims to resolve a timing issue related to the loading of fonts in the customize sidebar fonts panel. In certain scenarios, the panel was loading before the fonts were fully loaded, leading to a suboptimal user experience. To mitigate this issue, the following changes have been implemented:

Introducing Font Loading Check: Implemented a workaround using document.fonts.check to ensure that the custom sidebar fonts panel is loaded only after confirming the availability of the required fonts.

### DIFF
--- a/src/OnboardingSPA/components/Sidebar/components/Customize/DesignFontsPanel/index.js
+++ b/src/OnboardingSPA/components/Sidebar/components/Customize/DesignFontsPanel/index.js
@@ -262,6 +262,7 @@ const DesignFontsPanel = forwardRef(
 			bodySlug: style.font_content,
 		} ) );
 
+		const [ fontsLoaded, setFontsLoaded ] = useState( false );
 		const [ selectedGroup, setSelectedGroup ] = useState( null );
 		const [ showCustomFonts, setShowCustomFonts ] = useState( false );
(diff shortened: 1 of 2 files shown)


src/OnboardingSPA/components/Sidebar/components/Customize/DesignFontsPanel/index.js
```diff
@@ -337,6 +344,53 @@ const DesignFontsPanel = forwardRef(
...
+					setFontsLoaded( true );
+				}
+			}, 100 );
+
+			return () => {
+				clearInterval( checkFontsLoaded );
+			};
+			// eslint-disable-next-line react-hooks/exhaustive-deps
+		}, [ fontOptions ] );
+
```
arunshenoy99, line 383: @mr-vara we should avoid rendering hidden data unless absolutely necessary. The correct approach would be to wait for the fonts API to complete and then removing the skeleton. If the fonts are not loaded this seems a bigger issue with the Wonder Theme assets itself. Can we verify if this is necessary even after the expedite of the Wonder Theme installation and activation?

arunshenoy99 (changes requested):

PR #517, opened 2024-03-11
Install Plugins Extended

officiallygod:
## Proposed changes

Based on the Hiive Flags, a particular subset of plugins will be installed and activated.

## Type of Change

## Checklist

### DIFF
--- a/includes/Services/PluginService.php
+++ b/includes/Services/PluginService.php
@@ -64,7 +64,7 @@ class PluginService {
 			$init_plugins = array_merge( Plugins::get_init(), SiteFeatures::get_init() );
 		}
 
-		// Add plugins according to Hiive Flags
+		// Add plugins enabled by Hiive Flags
 		$init_plugins = array_merge( $init_plugins, self::initialize_hiive_flag_plugins() );
 
 		foreach ( $init_plugins as $init_plugin ) {
@@ -77,7 +77,7 @@ class PluginService {
 						new PluginInstallTask(
 							$init_plugin['slug'],
 							$init_plugin['activate'],
-							$init_plugin['priority']
+							isset( $init_plugin['priority'] ) ? $init_plugin['priority'] : 0
 						)
 					);
 					continue;
--- a/includes/Services/PluginService.php
+++ b/includes/Services/PluginService.php
@@ -22,28 +22,6 @@ use function NewfoldLabs\WP\ModuleLoader\container;
  * Class for providing plugin related services.
  */
 class PluginService {
-
-	/**
-	 * Gets the list of plugins based on different enabled Hiive Flags
-	 *
-	 * @return array
-	 */
-	public static function initialize_hiive_flag_plugins() {
-		$init_plugins_extended               = array();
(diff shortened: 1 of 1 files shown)


includes/Services/PluginService.php
```diff
@@ -21,6 +22,28 @@
...
+	 */
+	public static function initialize_hiive_flag_plugins() {
+		$init_plugins_extended = array();
+		$hiive_flags           = Plugins::get_hiive_plugin_flags();
+
+		foreach ( $hiive_flags as $hiive_flag => $is_enabled ) {
+			// Check if it is an allowed flag and is enabled on the container
+			if ( true === $is_enabled && true === Config::get_site_capability( $hiive_flag ) ) {
+				// Check if there are plugins for the flag.
+				if ( isset( Plugins::$hiive_flag_plugin_list[ $hiive_flag ] ) ) {
```
arunshenoy99, line 39: Can we check for a non-empty array instead of using `isset()`?

includes/Services/PluginService.php
```diff
@@ -21,6 +22,28 @@
...
 class PluginService {
+
+	/**
+	 * Gets the list of plugins based on different enabled Hiive Flags
+	 *
+	 * @return array
+	 */
+	public static function initialize_hiive_flag_plugins() {
+		$init_plugins_extended = array();
+		$hiive_flags           = Plugins::get_hiive_plugin_flags();
```
arunshenoy99, line 33: I think we can skip the extra `enablement` flag for the Hiive flags. I don't see a benefit in disabling them if the capability is already set. Essentially, we just move the plugins to a capability-to-plugins map and use it directly if the capability is set.

includes/Services/PluginService.php
```diff
@@ -41,6 +64,9 @@ public static function initialize() {
 			$init_plugins = array_merge( Plugins::get_init(), SiteFeatures::get_init() );
 		}
 
+		// Add plugins enabled by Hiive Flags
+		$init_plugins = array_merge( $init_plugins, self::initialize_hiive_flag_plugins() );
```
arunshenoy99, line 68: Do we still need to perform the plan checklist at https://github.com/newfold-labs/wp-module-onboarding-data/blob/a5e90dbdb748e2710f82654131b51973a6dabd12/includes/Plugins.php#L25? Is it necessary to verify the plan at https://github.com/newfold-labs/wp-module-onboarding-data/blob/a5e90dbdb748e2710f82654131b51973a6dabd12/includes/Plugins.php#L213? Can we simply load the plugins based on the capability? I'm just wondering if we can take this opportunity to simplify all of this complex logic.

arunshenoy99 (changes requested):

includes/Services/PluginService.php
officiallygod: Sure

includes/Services/PluginService.php
officiallygod: I will be updating this once I receive confirmation from the other end.

officiallygod: Done

arunshenoy99 (approved):

PR #519, opened 2024-03-12
Enhancement: Verify 'Yith Wonder' Theme Activation in Preview Step

mr-vara:
This pull request enhances the Preview Step in the SiteGen flow by introducing code to verify and ensure that the activated theme is 'Yith Wonder.' This addition aims to improve the reliability and consistency of the site generation process, ensuring that the correct theme is active during the preview.

### DIFF
--- a/src/OnboardingSPA/steps/SiteGen/Preview/index.js
+++ b/src/OnboardingSPA/steps/SiteGen/Preview/index.js
@@ -6,13 +6,24 @@ import { cloneDeep, isEmpty } from 'lodash';
 
 import CommonLayout from '../../../components/Layouts/Common';
 import { store as nfdOnboardingStore } from '../../../store';
-import { HEADER_SITEGEN } from '../../../../constants';
+import {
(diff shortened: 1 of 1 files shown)


arunshenoy99 (changes requested):
@mr-vara, I have some questions regarding this PR:
1. What happens to the in-progress API calls when the page is reloaded? Are they cancelled?
2. How are we handling the failure of the theme install API?
3. Why are we doing this only on the preview step? Should we use a StateHandler (preferably the `DesignStateHandler`) so that visiting any step that has a Live Preview will not result in errors?

The correct approach is to add `yith-wonder` as a prerequisite to the sitegen flow here: https://github.com/newfold-labs/wp-module-onboarding-data/blob/a5e90dbdb748e2710f82654131b51973a6dabd12/includes/Preview.php#L26 and then verify that instead of using the hardcoded `DESIGN_STEPS_THEME`. This will also ensure that the Wonder Theme is installed and activated on time, and we don't have to expedite every time.

PR #523, opened 2024-03-19
Enhancement: Verify 'Yith Wonder' Theme Activation in Preview & Editor Step

mr-vara:
This pull request enhances the Preview Step in the SiteGen flow by introducing code to verify and ensure that the activated theme is 'Yith Wonder.' This addition aims to improve the reliability and consistency of the site generation process, ensuring that the correct theme is active during the preview.

### DIFF
--- a/src/OnboardingSPA/steps/SiteGen/Preview/index.js
+++ b/src/OnboardingSPA/steps/SiteGen/Preview/index.js
@@ -6,7 +6,7 @@ import { cloneDeep, isEmpty } from 'lodash';
 
 import CommonLayout from '../../../components/Layouts/Common';
 import { store as nfdOnboardingStore } from '../../../store';
-import { HEADER_SITEGEN } from '../../../../constants';
+import { HEADER_SITEGEN, THEME_STATUS_ACTIVE } from '../../../../constants';
 import { SiteGenPreviewSelectableCard } from '../../../components/LivePreview';
 import getContents from './contents';
 import HeartAnimation from './heartAnimation';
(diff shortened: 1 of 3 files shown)


src/OnboardingSPA/steps/SiteGen/Preview/index.js
officiallygod, line 1: Just some Findings, So, `buildPreviews` checks for `homepages and globalStyles` which exist in this code
`if ( ! ( homepages && globalStyles ) || isPreviewLoading ) {`

but somehow the theme is not active due to which as we don't check the theme while we render it fails and crashes,
 
Screenshot 2024-04-01 at 4 43 56 PM

This could be a reason why Mike was getting those errors.

officiallygod (commented):
So the issue that is happening is expediteInstall is called and that installs and activates the yith theme and does a refresh of the same, 
and when that happens component loads again and `useEffect` is called that listens on `storedThemeStatus` as `storedThemeStatus = init` and handleThemeStatus is called and that reloads if theme was active due to which this useEffect cycle continues forever.

Looking for a fix and trying that.

officiallygod:
Onboarding Data PR: https://github.com/newfold-labs/wp-module-onboarding-data/pull/67

ajayadav09:
> * The previous flow continues to function as expected.
> * In the AI site generation flow, the YITH wonder is installed and activated. However, when customizing the color and typography, the typography changes are not applied.

if I applied a new font, it is still selecting the first font-pair from the Customize panel.

PR #532, opened 2024-04-03
A/B test for AI Site Generation

officiallygod:
## Proposed changes

Shows a Random different Fork Page to the user for A/B Testing

## Type of Change

## Checklist

## New Look
Screenshot 2024-04-05 at 12 12 35 PM
Screenshot 2024-04-05 at 12 13 51 PM
Screenshot 2024-04-05 at 1 49 26 PM
Screenshot 2024-04-05 at 1 50 12 PM

Note: Please check the earlier events if they are working as expected.


arunshenoy99 (commented):
@officiallygod Quick question: How does this work if we want to do A/B testing with other steps? Is it a good idea to use some kind of state handler that determines what needs to be rendered that we reuse?

officiallygod:
> @officiallygod Quick question: How does this work if we want to do A/B testing with other steps? Is it a good idea to use some kind of state handler that determines what needs to be rendered that we reuse?

The requirements are very specific to a section of the page and doing that would make us send all that data to the parent and again have a lot of dup code as we would then need to build 4 completely independent components which do the same thing (StartOptions in this case).

officiallygod:
Onboarding Data: https://github.com/newfold-labs/wp-module-onboarding-data/pull/68

arunshenoy99 (approved):

PR #541, opened 2024-04-15
Add Defensive Checks Just in case

officiallygod:
## Type of Change

## Checklist

## Bug

Screenshot 2024-04-05 at 8 06 36 AM

After spending time digging into the problem, I figured out why we were having issues. It turns out that the data we were getting back for the step header-menu wasn't complete. Basically, it was missing the page header, which caused a JavaScript error.

Normally, the header-menu API gives us default `wonder-blocks` headers. But if those are missing, it switches to headers from `Yith Wonder`. And now, if the theme doesn't exist, we've fixed it so a modal pops up instead, which was introduced in the latest BH release. Earlier, this modal wasn't showing up, but now it is.

To prevent any more errors on the screen, I've added extra checks and fallbacks. These are just extra precautions to make sure everything runs smoothly and users don't encounter any problems.


arunshenoy99 (commented):
@officiallygod, do you think it would be good to show some kind of message to the user in this edge case when Wonder won't be active?

officiallygod:
> @officiallygod, do you think it would be good to show some kind of message to the user in this edge case when Wonder won't be active?

Ohhh @arunshenoy99 that is a good question, Let me clarify that with someone, Thanks.

arunshenoy99 (approved):

PR #565, opened 2024-05-20
Press0 1328 default to light mode

ajayadav09:
## Proposed changes

Default to light mode unless the user has prefer-color-scheme as dark mode. 
## Type of Change

## Checklist


arunshenoy99 (commented):
@ajayadav09, can we explore using `useEffect` instead of an event listener, and can we fix the failing tests? Thanks!

ajayadav09:
> @ajayadav09, can we explore using `useEffect` instead of an event listener, and can we fix the failing tests? Thanks!

@arunshenoy99  I tried the useEffect approach to check for changes but the window.matchMedia is a function provided by the window object and it does not change between renders. So to check for changes will have to add listeners to matchMedia.

arunshenoy99 (approved):

PR #575, opened 2024-06-06
Fix Multilingual Reporting of Event Values

arunshenoy99:
## Proposed changes
Fix the reporting of event values in both English and Brazilian Portuguese for the following events:
1. Onboarding Step Skipped (moving from internationalized step names to slugs). For example, 'Basic Info' would become 'basic-info,' and similarly, every step now has its own slug that will not be translated.
2. Onboarding Starter Pages Selected (moving from internationalized page names to slugs). For example, 'About' would become 'about,' and similarly, every Site Page has its own key.

## Type of Change

## Checklist

## Further comments
Goes with: https://github.com/newfold-labs/wp-module-onboarding-data/pull/95


officiallygod (approved):
Looks Good

PR #583, opened 2024-06-17
Press0 1453 dark mode preview

ajayadav09:
## Type of Change

## Checklist

### DIFF
--- a/src/OnboardingSPA/components/Sidebar/components/Customize/DesignFontsPanel/index.js
+++ b/src/OnboardingSPA/components/Sidebar/components/Customize/DesignFontsPanel/index.js
@@ -356,6 +356,7 @@ const DesignFontsPanel = forwardRef(
 
 		const handleSelectYourOwnFonts = () => {
 			setShowCustomFonts( true );
+			setSelectedGroup( 'custom' );
 			if ( ! selectedCustomFont ) {
 				setIsEditingCustomFont( true );
 			}
@@ -366,7 +367,7 @@ const DesignFontsPanel = forwardRef(
 		};
 
 		const handleCancelCustomFonts = () => {
-			if ( ! selectedCustomFont ) {
+			if ( ! selectedCustomFont.heading || ! selectedCustomFont.body ) {
 				setShowCustomFonts( false );
 			} else {
 				setIsEditingCustomFont( false );


officiallygod (commented):
This looks good but should we clear the localStorage values created during onboarding? As, I suppose these might not be useful later and might take up memory.

Also, considering we use strings like `'nfd-sitegen-theme-mode'` very often should we go about using constants? or helper methods to set data or retrieve data like we do using **Options** and **get_option_name** in Php?

ajayadav09:
> This looks good but should we clear the localStorage values created during onboarding? As, I suppose these might not be useful later and might take up memory.
> 
> Also, considering we use strings like `'nfd-sitegen-theme-mode'` very often should we go about using constants? or helper methods to set data or retrieve data like we do using **Options** and **get_option_name** in Php?

@officiallygod  @arunshenoy99  moved from localstorage to store and saving the theme value in onboarding data here

https://github.com/newfold-labs/wp-module-onboarding-data/pull/101

src/OnboardingSPA/components/ThemeContextProvider/index.js
```diff
@@ -1,28 +1,54 @@
...
+	const { currentData } = useSelect( ( select ) => {
+		return {
+			currentData:
+				select( nfdOnboardingStore ).getCurrentOnboardingData(),
+		};
+	} );
+
+	if ( currentData.sitegenThemeMode ) {
+		return currentData.sitegenThemeMode;
+	}
```
officiallygod, line 19: If this is '' does it fall here? Cause ig the default theme is the Device theme

src/OnboardingSPA/components/ThemeContextProvider/index.js
```diff
@@ -1,28 +1,54 @@
...
 	) {
 		return THEME_DARK;
 	}
+
 	return THEME_LIGHT;
 }
 
 const ThemeProvider = ( { children } ) => {
-	const [ theme, setTheme ] = useState( getPreferredColorScheme );
+	const preferredColorScheme = usePreferredColorScheme();
```
officiallygod, line 32: Considering here we have to call **usePreferredColorScheme** does it make sense to initialize `currentData` here and as well as in the above func or send as an param to the func

src/OnboardingSPA/components/ThemeContextProvider/index.js
```diff
@@ -1,28 +1,54 @@
...
 		const colorSchemeMediaQuery = window.matchMedia(
 			'(prefers-color-scheme: dark)'
 		);
 
 		const handleChange = ( event ) => {
-			setTheme( event.matches ? THEME_DARK : THEME_LIGHT );
+			const newTheme = event.matches ? THEME_DARK : THEME_LIGHT;
+			setTheme( newTheme );
+			currentData.sitegenThemeMode = newTheme;
+			setFlow( currentData );
```
officiallygod, line 51: Considering we are changing the data shouldn't we also do `setCurrentOnboardingData` and sync it to the store?

src/OnboardingSPA/components/ThemeContextProvider/index.js
```diff
@@ -33,9 +59,13 @@ const ThemeProvider = ( { children } ) => {
 	}, [] );
 
 	const toggleTheme = () => {
-		setTheme( ( prevTheme ) =>
-			prevTheme === THEME_DARK ? THEME_LIGHT : THEME_DARK
-		);
+		setTheme( ( prevTheme ) => {
+			const newTheme =
+				prevTheme === THEME_DARK ? THEME_LIGHT : THEME_DARK;
+			currentData.sitegenThemeMode = newTheme;
```
officiallygod, line 65: `setCurrentOnboardingData`

src/OnboardingSPA/store/reducer.js
```diff
@@ -35,6 +35,7 @@ export function flow(
 		},
 		chapter: undefined,
 		interactionDisabled: false,
+		sitegenThemeMode: '',
```
officiallygod, line 38: I doubt if this is required as we always have the data in currentData.sitegenThemeMode, then why here?

src/OnboardingSPA/components/ThemeContextProvider/index.js
ajayadav09: if the user toggles the theme, currentData.sitegenThemeMode will be set to that and if the user refreshes the page or revisits, the last theme will be applied.

officiallygod (approved):

arunshenoy99:
PRESS0-1453

PR #587, opened 2024-06-24
Remove `flow` param for Events API requests

arunshenoy99:
## Proposed changes

For the events API, since we are initializing the namespace at the beginning of onboarding, we will not be able to keep up with the change in the flow parameter, which is part of the URL. This parameter is not being used at all in the events API and is also unnecessary.

## Type of Change

## Checklist

- [] I have added tests that prove my fix is effective or that my feature works


ajayadav09 (approved):

arunshenoy99:
PRESS0-1587

PR #614, opened 2024-08-29
Fix Failing Onboarding Test

officiallygod:
## Proposed changes

Investigate and fix failing test cases.
The navigation back button should not exist when the Capability hasAISitegen is false however now we set hasAISitegen by default for users in [Link](https://github.com/newfold-labs/wp-module-data/blob/91d1b03162f23b7b8bd3a41d9496eb8017df0051/bootstrap.php#L79-L92) and fork is always enabled even in the cypress env due to which the Back button exists.

Solution:

1. Update Cypress to reverse the check, and when the hardcoded hasAISitegen is removed again, change it to the current state.As the hardcoded changes only exist in the Bluehost plugin, opting for this solution might also require brand-specific checks
2. Remove this check itself as AI Sitegen happens for 100% of customers and there is no case where there will be no fork page.

Going with approach 2 and removing this as doing all those checks is an overkill.

## Type of Change

## Checklist

- [x] I have viewed my change in a web-browser

### DIFF
--- a/src/OnboardingSPA/index.js
+++ b/src/OnboardingSPA/index.js
@@ -50,6 +50,7 @@ export async function initializeNFDOnboarding( id, runtime ) {
 	}
 
 	const currentData = await getFlow();
+	
 	if ( currentData.error === null ) {
 		currentData.body = initializeFlowData( currentData.body );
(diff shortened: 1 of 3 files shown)


officiallygod:
PRESS0-2171

wpscholar (approved):

PR #632, opened 2024-11-04
Redirect interceptor

wpalani:
## Proposed changes
Filter redirect calls while on the onboarding page.

More info: https://jira.newfold.com/browse/PRESS8-49

## Type of Change

## Checklist

### DIFF
--- a/includes/ExternalRedirectInterceptor.php
+++ b/includes/ExternalRedirectInterceptor.php
@@ -1,0 +1,38 @@
+<?php
+namespace NewfoldLabs\WP\Module\Onboarding;
+
+use NewfoldLabs\WP\Module\Onboarding\Data\Data;
+
+/**
+ * Class to intercept redirect calls and filter them.
+ *
+ * The purpose of this class is to prevent any redirects while the user is on the onboarding page.
+ * The only allowed redirect is to the brand plugin page.
+ */
+class ExternalRedirectInterceptor {
+	public function __construct() {
+		if ( ! isset( $_GET['page'] ) || WP_Admin::$slug !== \sanitize_text_field( $_GET['page'] ) ) {
+			return;
+		}
+
+		\add_filter( 'wp_redirect', array( $this, 'wp_redirect' ), 10, 1 );
+	}
+
+	/**
+	 * Intercept wp_redirect calls and filter them.
+	 *
+	 * @param string $location The location to redirect to.
+	 */
+	public function wp_redirect($location): string {
+		$runtime_data = Data::runtime();
+		$brand_plugin_url  = $runtime_data['currentBrand']['pluginDashboardPage'];
+
+		// Intercept if the redirect is going anywhere other than the brand plugin page.
+		if ( strpos( $location, $brand_plugin_url ) !== 0 ) {
+			return '';
+		}
+
+		// Allow the redirect if it's going to the brand plugin page.
+		return $location;
+	}
+}
--- a/includes/ExternalRedirectInterceptor.php
+++ b/includes/ExternalRedirectInterceptor.php
@@ -10,8 +10,11 @@ use NewfoldLabs\WP\Module\Onboarding\Data\Data;
  * The only allowed redirect is to the brand plugin page.
  */
 class ExternalRedirectInterceptor {
+	/**
+	 * Constructor.
+	 */
 	public function __construct() {
-		if ( ! isset( $_GET['page'] ) || WP_Admin::$slug !== \sanitize_text_field( $_GET['page'] ) ) {
+		if ( ! isset( $_GET['page'] ) || \sanitize_text_field( $_GET['page'] ) !== WP_Admin::$slug ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


includes/ExternalRedirectInterceptor.php
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	/**
+	 * Intercept wp_redirect calls and filter them.
+	 *
+	 * @param string $location The location to redirect to.
+	 */
+	public function wp_redirect( $location ): string {
+		$runtime_data = Data::runtime();
+		$brand_plugin_url  = $runtime_data['currentBrand']['pluginDashboardPage'];
```
wpscholar, line 31: @wpalani What happens if one of these array keys doesn't exist?

includes/ExternalRedirectInterceptor.php
```diff
@@ -0,0 +1,41 @@
...
+
+		// Intercept if the redirect is going anywhere other than the brand plugin page.
+		if ( strpos( $location, $brand_plugin_url ) !== 0 ) {
```
wpscholar, line 34: @wpalani Is the logic here right? If not found, it returns false and since `false !== 0`, it evaluates to true. If found, but not at the beginning, then it evaluates to a number greater than 0.

includes/ExternalRedirectInterceptor.php
wpalani: The runtime should always return a value for both `currentBrand` and `pluginDashboardPage`

https://github.com/newfold-labs/wp-module-onboarding-data/blob/main/includes/Data.php#L43-L49

wpscholar: Yes, but given the brands defined: https://github.com/newfold-labs/wp-module-onboarding-data/blob/main/includes/Brands.php#L44-L471

And these HostGator locales: https://github.com/newfold-labs/wp-plugin-hostgator/blob/main/src/app/data/region.js#L9-L18

I'm wondering if it might still happen.

includes/ExternalRedirectInterceptor.php
wpalani: Yes, it's correct; if not found and the value is `false`, we shouldn't redirect. and if found and but the position is not `0` we shouldn't redirect either. Here's a breakdown of the logic:

1. If `$brand_plugin_url` is found at the beginning of `$location`, `strpos($location, $brand_plugin_url)` will return 0, which satisfies !== 0 as `false`. Therefore, it won’t enter the if block, allowing the redirect.
2. If `$brand_plugin_url` is not found at all, `strpos()` will return `false`, and false !== 0 evaluates to true, so the if condition is met, returning `''` and preventing the redirect.
3. If `$brand_plugin_url` is found but not at the beginning, `strpos()` will return a value greater than 0, which also satisfies the !== 0 check, returning `''` and stopping the redirect.

I can def see how this line can be hard to read though (to you point), so I just pushed this small code change to make it more readable:

```php
public function wp_redirect( $location ): string {
  $runtime_data     = Data::runtime();
  $brand_plugin_url = $runtime_data['currentBrand']['pluginDashboardPage'];
  
  $location_is_brand_plugin_url = strpos( $location, $brand_plugin_url );
  // Intercept if the redirect is going anywhere other than the brand plugin page.
  if ( false === $location_is_brand_plugin_url || 0 !== $location_is_brand_plugin_url ) {
	  return '';
  }
  
  // Allow the redirect if it's going to the brand plugin page.
  return $location;
}
```

includes/ExternalRedirectInterceptor.php
wpalani: If this may happen for smaller brands, we can be more defensive here since it's not a critical path feature (more of a firewall). I'll push a code to check the values gracefully before processing them.

wpalani: Here's the commit: https://github.com/newfold-labs/wp-module-onboarding/pull/632/commits/ca5e6533f31c23ef81fd81fb446a91ff036623b0

includes/ExternalRedirectInterceptor.php
```diff
@@ -0,0 +1,65 @@
...
+		 */
+		// Check if the current brand is set and is an array.
+		if (
+			isset( $runtime_data['currentBrand'] ) &&
+			is_array( $runtime_data['currentBrand'] )
+			) {
+			// Check if the pluginDashboardPage key is set and is a string.
+			if (
+				isset( $runtime_data['currentBrand']['pluginDashboardPage'] ) &&
+				is_string( $runtime_data['currentBrand']['pluginDashboardPage'] )
```
wpscholar, line 44:
```suggestion
			isset( $runtime_data['currentBrand'], $runtime_data['currentBrand']['pluginDashboardPage'] ) &&
			is_string( $runtime_data['currentBrand']['pluginDashboardPage'] )
```
@wpalani  We can probably simplify to this

amartya-dev (approved):
looks good

PR #819, opened 2025-07-24
saving site info in options table

abhijitb:
## Proposed changes

This PR  has changes that will create a database option with serialized array type "nfd_module_onboarding_site_info" when Onboarding completes so that other modules can access it.

example data :
```
{ "experience_level": 3,  "site_type" : "business" }
```

#### Production

- [ ] Dependency update

#### Development

- [ ] Tests
- [ ] Dependency update
- [ ] Documentation Update

## Checklist

### DIFF
--- a/includes/Services/AppService.php
+++ b/includes/Services/AppService.php
@@ -6,6 +6,7 @@ use NewfoldLabs\WP\Module\Onboarding\Data\Events;
 use NewfoldLabs\WP\Module\Onboarding\Data\Options;
 use NewfoldLabs\WP\Module\Onboarding\Data\Services\PreviewsService;
 use NewfoldLabs\WP\Module\Onboarding\Data\Services\SiteGenService as LegacySiteGenService;
+use NewfoldLabs\WP\Module\Onboarding\Data\Services\FlowService;
 
 
 use function NewfoldLabs\WP\ModuleLoader\container;
@@ -58,6 +59,9 @@ class AppService {
 
 		// Mark onboarding as completed.
 		StatusService::handle_completed();
+
+		// Save onboarding site information for other modules to access.
+		SiteInfoService::save_site_info();
 
 		// Purge all caches.
 		container()->get( 'cachePurger' )->purge_all();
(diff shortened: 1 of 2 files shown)


wpalani (changes requested):
- [x] I think a more appropriate place add the code is in here: https://github.com/newfold-labs/wp-module-onboarding/blob/trunk/includes/Services/StatusService.php#L47-L56
- [x]  I'm not sure we need service class for this, especially to prevent other modules from relying on Onboarding directly. A simple method in the `StatusService` class to inject the option in the db is enough. And then other modules can fetch the data from the DB.
- [x] We no longer use the FlowService, so this code won't work. Instead, you can use the `ReduxStateService` to access the data from Onboarding, here's a usage example: https://github.com/newfold-labs/wp-module-onboarding-data/blob/main/includes/Services/FlowService.php#L439-L441

I am open to the idea of adding a new class that include getter methods of the onboarding data, especially the input slice or adding these methods to the existing ReduxService class.

abhijitb:
@wpalani  I have made the necessary changes. Please have a look whenever you can.. thanks.

includes/Services/StatusService.php
```diff
@@ -181,11 +186,28 @@ public static function track(): void {
...
+	 * Save onboarding site information to database option for other modules to access.
+	 *
+	 * @return void
+	 */
+	public static function save_site_info(): void {
+		$site_info = array();
+
+		// Get experience level and site type from ReduxStateService
+		$data = ReduxStateService::get( 'input' );
+		$site_info['experience_level'] = $data['experienceLevel'] ?? 3;
```
wpalani, line 207: I see that we're falling back on `int` 3 but the `experienceLevel` is stored as a string, see how it's being translated here: https://github.com/newfold-labs/wp-module-onboarding-data/blob/main/includes/Services/FlowService.php#L442-L453

However, that being said, i'd rather store it as string since that's more readable, so you'll just have fallback to 'advanced' instead of number 3.

PR #855, opened 2025-08-11
Refine the options passed to the `make-pot` command

desrosj:
## Proposed changes

This makes the following change to the `make-pot` command:
- explicitly pass `--domain`: The default value for `--domain` is the project slug when `Text Domain` is not specified in the main plugin or theme file. Explicitly passing the text domain avoids the scenario where someone checks the repository out in a folder with a custom name.

#### Production

- [ ] Dependency update

#### Development

- [ ] Tests
- [ ] Dependency update
- [ ] Documentation Update

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
 		"fix": [
 			"vendor/bin/phpcbf . --standard=phpcs.xml"
 		],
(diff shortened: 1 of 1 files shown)


wpscholar (approved):

circlecube:
> @circlecube - Can you look at why the tests aren't running here?

New onboarding doesn't have tests yet, so we skip onboarding tests in the bluehost plugin. The other plugins are running old onboarding tests on the new build, so they're failing. Need to have tests written for the new onboarding, and we can remove it from the exclude list in bluehost.

We were told to hold off on updating onboarding in other brands, but once tests are updated, theoretically, they should pass in all brand plugins as long as there is no brand-specific logic stopping pages from loading.

PR #933, opened 2025-09-29
Navigate to blueprints on AI sitegen fail

zlateska:
## Proposed changes

This PR updates the AI SiteGen flow to navigate users to the Blueprints screen when generation fails. It also displays a custom error message to inform users about the failure and guide them on next steps.

#### Production

- [ ] Dependency update

#### Development

- [ ] Tests
- [ ] Dependency update
- [ ] Documentation Update

## Checklist

### DIFF
--- a/src/app/steps/Blueprints/BlueprintsStep.js
+++ b/src/app/steps/Blueprints/BlueprintsStep.js
@@ -211,13 +211,15 @@ const BlueprintsStep = () => {
 					) }
 				</Container.Block>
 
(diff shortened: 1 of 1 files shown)


wpalani (changes requested):
One thing stood out to me: instead of passing route context we can just check the store state `sitegenHasFailed` on the blueprints page and render a different title.

that way the components are not dependent on each other directly and the store is the source of truth.

PR #1146, opened 2026-04-09
Onboarding: SiteGen published/selected logic

kreamweb:
## Proposed changes

https://newfold.atlassian.net/browse/PRESS0-4233
https://newfold.atlassian.net/browse/PRESS0-4218

Adds **SiteGenServiceRequest** service class to handle **authenticated**, async HTTP requests to the AI Platform.

Adds **SiteGenAiController** with two REST endpoints:

`POST /sitegen/handshake` — performs a handshake with the AI Platform, returning the site_id for the current site
`POST /sitegen/report-published` — reads the stored `sitegen_current_id` and fires a fire-and-forget request to the AI Platform (sitegen/select) to report the published site

Registers all **SiteGen Options** keys in Options.php:

`sitegen_site_id` — the site identifier returned by the AI Platform on handshake
`sitegen_current_id` — the most recent SiteGen generation ID
`sitegen_previous_ids` — array of all previous SiteGen IDs (accumulated on each new generation)
`sitegen_site_type` — the site type determined during discovery
`sitegen_enhanced_prompt` — the enhanced prompt produced by the AI
`sitegen_discovery_data` — full discovery results, stored for potential post-onboarding use

**Each field of the sitegen Redux slice is now persisted in its own dedicated WordPress option** instead of a single serialised blob. 
`ReduxStateController::update_sitegen_slice_state()` writes each field individually; `get_sitegen_slice_state()` reconstructs the slice from those options. 
`ReduxStateService::get('sitegen')` uses the same logic (via get_state_sitegen()) so the bootstrap data inlined on page load is also sourced from the individual options.

When a new `sitegen_current_id` arrives and differs from the stored value, the old ID is appended to `sitegen_previous_ids` before the current ID is updated.

Calls `reportSiteGenPublished()` at the end of the publish pipeline, after onboarding completes and the site is live.

**Note**: Two options intentionally survive an onboarding restart:
`sitegen_site_id` — the site identity does not change between runs; the handshake is not repeated
`sitegen_previous_ids` — preserves the history of all past generation IDs across restarts

**Important**: `sitegen_site_id` is tied to the domain registered with the AI Platform. When switching environments (e.g. local → staging → production) the stored `site_id` will not be recognised by the Platform. In that case the option `nfd_module_onboarding_sitegen_site_id` should be removed manually from the database before running the onboarding again.

#### Production

- [ ] Dependency update

#### Development

- [ ] Tests
- [ ] Dependency update
- [ ] Documentation Update

## Checklist

### DIFF
--- a/src/app/hooks/useChat.js
+++ b/src/app/hooks/useChat.js
@@ -2,6 +2,8 @@
  * WordPress dependencies
  */
 import { useCallback, useState, useRef, useEffect } from '@wordpress/element';
+import { dispatch, select } from '@wordpress/data';
+import { nfdOnboardingStore } from '@/data/store';
 
 /**
  * Internal dependencies
@@ -26,7 +28,6 @@ import {
 	createInitialTasks,
 	createGenerationTasks,
 } from '@/hooks/chat/tasks';
-import { saveSitegenOptions } from '@/utils/api/onboarding';
 
 /**
  * useChat hook
@@ -107,7 +108,7 @@ const useChat = () => {
 
 				await streamSSE( response, ( { event, data } ) => {
 					if ( event === 'sitegen_started' ) {
-						saveSitegenOptions( { sitegen_id: data } );
+						dispatch( nfdOnboardingStore ).setSiteGenId( data );
 						return;
 					}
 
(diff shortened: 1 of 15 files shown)


src/app/hooks/useChat.js
```diff
@@ -116,6 +121,7 @@ const useChat = () => {
 					if ( discoveryTaskKey ) {
 						if (
 							discoveryTaskKey === 'brand_identity' ||
+							discoveryTaskKey === 'sitemap' ||
```
wpalani, line 135: I think we've removed the `sitemap` task, because the AI service no longer expose that.

kreamweb:
I moved to draft this PR, in this https://newfold.atlassian.net/browse/PRESS0-4218, the site_id and sitegen_id, will be stored in a different way.

wpalani (approved):

PR #1181, opened 2026-05-13
Support svg

kreamweb:
## Proposed changes

This filter to fix again the icon issue, that often are empty.
WordPress sanitizes block content and block attributes that store HTML through wp_kses (typically with the post allowlist). Core’s default allowed tags for that context do not include SVG elements (svg, path, rect, etc.). On save, filter_block_content / filter_block_kses strips those tags, so the icon appears empty on the front end and in reloaded editor content.

#### Production

- [ ] Dependency update

#### Development

- [ ] Tests
- [ ] Dependency update
- [ ] Documentation Update

## Checklist

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -1,7 +1,6 @@
 <?php
 
 use NewfoldLabs\WP\ModuleLoader\Container;
-use NewfoldLabs\WP\Module\Onboarding\Application;
 use NewfoldLabs\WP\Module\Onboarding\ModuleController;
 use NewfoldLabs\WP\Module\Onboarding\Compatibility\Scan;
 use NewfoldLabs\WP\Module\Onboarding\Compatibility\Safe_Mode;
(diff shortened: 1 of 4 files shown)


wpalani (approved):

kreamweb:
@wpalani I updated the code:

The SVG KSES filter (`wp_kses_allowed_html`) was only registered inside `WP_Admin::__construct(),` which is instantiated exclusively in authenticated admin contexts. 
WP-Cron runs as a separate non-blocking HTTP request with no logged-in user, so `WP_Admin` is never instantiated and the filter is never registered.

As a result, when the cron job called `wp_update_post()` to rewrite external image URLs in post content, WordPress applied `wp_kses_post()` without SVG support, stripping all inline SVG tags (<svg>, <path>, <circle>, etc.) from the page content.

The fix registers the filter directly inside `update_post_content_with_new_image_urls()`, immediately before wp_update_post() `is` called, and removes it afterwards. This ensures SVG tags are preserved regardless of the execution context (admin, REST, or cron).


Added also a **daily re-scan** of onboarding pages to sideload images replaced after onboarding

PR #1224, opened 2026-06-29
Add blog category navigation, homepage sections, and archive template

zlateska:
When we generate a blog site, the homepage and menu should actually reflect the blog's categories. This PR makes that happen.

A sitekit can now ship a category archive page (the page you see at `/category/news/`), and we apply it during publish. We also start tracking which generated page is the home page and which is the contact page, since the menu needs to know.

Once the articles are published, their categories exist, so we go back to the homepage and fill in its blog sections — each one gets a real category, so it shows the right posts and the right heading and "Read more" link.

Finally, we add the top categories to the main menu and keep the menu to five items, ordered home → categories → other pages → contact, so the categories don't get pushed out. If the header button already links to Contact, we leave Contact out of the menu so it's not there twice.

Also fixes a small existing typo in the nav code (`$$navigation`).

### DIFF
--- a/includes/Services/SiteTypes/BlogSiteTypeService.php
+++ b/includes/Services/SiteTypes/BlogSiteTypeService.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Resolver for category-bound blog sections.
+ * Blog site type service.
  *
  * @package NewfoldLabs\WP\Module\Onboarding\Services\SiteTypes
  */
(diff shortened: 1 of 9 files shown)


includes/Services/SiteTypes/BlogSiteTypeService.php
wpalani, line 1: It feels this class should live in the `SiteTypes` namespace, it's more of a helper class to the CommonSiteTypeService.

We either make a new `BlogSiteTypeService` class and have it live there or move it a level up to the `Services` namespace

wpalani (dismissed):
Small change request. but the rest looks good.

wpalani (approved):

